PR #65, opened 2020-12-03
now we can get tables names defined in the view, and remove a extra c…

MartinSun34:
…omment symbol

To get the table names defined in VIEWS, I moved the 'get table name' part from post_analyze_hook to the function after rewritten.


ibrarahmad:
Please rebase that, it has conflitct.

PR #66, opened 2020-12-23
Add placeholder replacement support for prepared statement.

MartinSun34:
What I did is when user set normalized_query to off, pg_stat_monitor will store a query with actual parameter as the query_text.
Here is simple example, assume we have table like following
```
test_stat_monitor=# \d employee
              Table "public.employee"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 id     | integer |           |          | 
 name   | text    |           |          | 

test_stat_monitor=# select * from employee;
 id |  name  
----+--------
  1 | kevin
  2 | stuart
  3 | bob
(3 rows)
```
And if we execute a parpared statement we will reuslt like this: 
```
test_stat_monitor=# select pg_stat_monitor_reset();
 pg_stat_monitor_reset 
-----------------------
 
(1 row)

test_stat_monitor=# prepare getemp as select * from employee where id = $1 and name = $2;
PREPARE
test_stat_monitor=# execute getemp(3,'bob');
 id | name 
----+------
  3 | bob
(1 row)

test_stat_monitor=# select query,calls from pg_stat_monitor;
                                query                                 | calls 
----------------------------------------------------------------------+-------
 select pg_stat_monitor_reset()                                       |     1
 prepare getemp as select * from employee where id = $1 and name = $2 |     1
(2 rows)
```
Even we set normalized_query to off we will still see the same result. Because for a prepared statement pgss_ExecutorEnd will try to get query text from queryDesc->sourceText, it have placeholder in it. I add a funciton can replace placeholder symbol with actual parameters. if user set normalized_query to off the repalcement function will be called. Now we can have the following result:

```
test_stat_monitor=# select pg_stat_monitor_reset();
 pg_stat_monitor_reset 
-----------------------
 
(1 row)

test_stat_monitor=# set pg_stat_monitor.pgsm_normalized_query to off;
SET
test_stat_monitor=# execute getemp(3,'bob');
 id | name 
----+------
  3 | bob
(1 row)

test_stat_monitor=# select query,calls from pg_stat_monitor;
                                query                                 | calls 
----------------------------------------------------------------------+-------
 prepare getemp as select * from employee where id = 3 and name = bob |     1
 select pg_stat_monitor_reset()                                       |     1
 set pg_stat_monitor.pgsm_normalized_query to off                     |     1
(3 rows)
```

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -35,8 +35,6 @@ static int	exec_nested_level = 0;
 static bool system_init = false;
 static struct rusage  rusage_start;
 static struct rusage  rusage_end;
-static volatile sig_atomic_t sigterm = false;
-static void handle_sigterm(SIGNAL_ARGS);
 static unsigned char *pgss_qbuf[MAX_BUCKETS];
 
 
@@ -59,7 +57,6 @@ PG_FUNCTION_INFO_V1(pg_stat_monitor_reset);
 PG_FUNCTION_INFO_V1(pg_stat_monitor_1_2);
 PG_FUNCTION_INFO_V1(pg_stat_monitor_1_3);
 PG_FUNCTION_INFO_V1(pg_stat_monitor);
-PG_FUNCTION_INFO_V1(pg_stat_wait_events);
 PG_FUNCTION_INFO_V1(pg_stat_monitor_settings);
 
 static uint pg_get_client_addr(void);
@@ -130,10 +127,6 @@ static uint64 get_next_wbucket(pgssSharedState *pgss);
 static void store_query(uint64 queryid, const char *query, uint64 query_len);
 static uint64 locate_query(uint64 bucket_id, uint64 queryid, char * query);
 
-/* Wait Event Local Functions */
-static void register_wait_event(void);
-void wait_event_main(Datum main_arg);
-static void update_wait_event(void);
 static uint64 get_query_id(pgssJumbleState *jstate, Query *query);
 
 /*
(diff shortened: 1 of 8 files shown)


pg_stat_monitor.c
```diff
@@ -409,9 +413,18 @@ pgss_ExecutorEnd(QueryDesc *queryDesc)
 		utime = TIMEVAL_DIFF(rusage_start.ru_utime, rusage_end.ru_utime);
 		stime = TIMEVAL_DIFF(rusage_start.ru_stime, rusage_end.ru_stime);
 
+		/* 
+		 * For a query with parameters if we want the actual query we denormalized it
+		 * by replacing all the placeholders with actual parameter values.
+		 */
+        if(!PGSM_NORMALIZED_QUERY && queryDesc->params)
```
grantzhou, line 420: Use tabs to indent the start of the line,  follow the original coding format

ibrarahmad:
Thanks for working on this. You need to perform to make that patch(pull request) reviewed and committed.
1  - Create a feature/bug request here (https://jira.percona.com/projects/PG/issues).
2 - Apply your patch with the only commit by following these rules.
     2.1 - Pull request must contain one complete commit.
     2.2 - Commit message should contain Jira ticket number followed by a single line description and then full stop.
     2.3 - One blank line.
     2.4 - A detailed description of the feature/problem. 
     2.5 - Example.
---
                PG-150: Logging CMD Type like SELECT, UPDATE, INSERT, DELETE.
                
                Detail if needed.
                
                Reviewed By: XYZ.
---

  3 - Push your code.
  4 - Create a pull request.
              

PS: Don't forget to change the README file, if needed.

PR #67, opened 2021-01-04
PG-156: Adding a placeholder replacement function for prepared statement

MartinSun34:
What I did is when user set normalized_query to off, pg_stat_monitor will store a query with actual parameter as the query_text.
Here is simple example, assume we have table like following
```
test_stat_monitor=# \d employee
              Table "public.employee"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 id     | integer |           |          | 
 name   | text    |           |          | 

test_stat_monitor=# select * from employee;
 id |  name  
----+--------
  1 | kevin
  2 | stuart
  3 | bob
(3 rows)
```
And if we execute a parpared statement we will reuslt like this: 
```
test_stat_monitor=# select pg_stat_monitor_reset();
 pg_stat_monitor_reset 
-----------------------
 
(1 row)

test_stat_monitor=# prepare getemp as select * from employee where id = $1 and name = $2;
PREPARE
test_stat_monitor=# execute getemp(3,'bob');
 id | name 
----+------
  3 | bob
(1 row)

test_stat_monitor=# select query,calls from pg_stat_monitor;
                                query                                 | calls 
----------------------------------------------------------------------+-------
 select pg_stat_monitor_reset()                                       |     1
 prepare getemp as select * from employee where id = $1 and name = $2 |     1
(2 rows)
```
Even we set normalized_query to off we will still see the same result. Because for a prepared statement pgss_ExecutorEnd will try to get query text from queryDesc->sourceText, it have placeholder in it. I add a funciton can replace placeholder symbol with actual parameters. if user set normalized_query to off the repalcement function will be called. Now we can have the following result:

```
test_stat_monitor=# select pg_stat_monitor_reset();
 pg_stat_monitor_reset 
-----------------------
 
(1 row)

test_stat_monitor=# set pg_stat_monitor.pgsm_normalized_query to off;
SET
test_stat_monitor=# execute getemp(3,'bob');
 id | name 
----+------
  3 | bob
(1 row)

test_stat_monitor=# select query,calls from pg_stat_monitor;
                                query                                 | calls 
----------------------------------------------------------------------+-------
 prepare getemp as select * from employee where id = 3 and name = bob |     1
 select pg_stat_monitor_reset()                                       |     1
 set pg_stat_monitor.pgsm_normalized_query to off                     |     1
(3 rows)
```

Reviewd By: Hamid Akhtar

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -245,6 +245,8 @@ pg_stat_monitor_version(PG_FUNCTION_ARGS)
 	PG_RETURN_TEXT_P(cstring_to_text(BUILD_VERSION));
 }
 
+static char **get_params_text_list(const ParamListInfo paramlist);
+static char *get_denormalized_query(const ParamListInfo paramlist, const char *query_text);
 
 /*
  * Post-parse-analysis hook: mark query with a queryId
@@ -397,6 +399,7 @@ pgss_ExecutorEnd(QueryDesc *queryDesc)
 	float   utime;
 	float   stime;
 	uint64	queryId = queryDesc->plannedstmt->queryId;
+	const char  *query_text;
 
 	if (queryId != UINT64CONST(0) && queryDesc->totaltime)
 	{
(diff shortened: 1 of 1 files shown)


EngineeredVirus (commented):
A couple of feedbacks on your patch:
(1) palloc0 followed by a memset for '\0' is redundant. From a performance perspective, I would suggest to use palloc instead and keep the second memset statement setting '\0'.
(2) I think the loop that's scanning through the string may have an issue when '$' sign is part of the query that's not a placeholder. 
- This needs further testing though

ibrarahmad:
I reverted the patch because it breaks some existing functionality.

```
postgres=# select query,calls from pg_stat_monitor;
                                 query                                 | calls 
-----------------------------------------------------------------------+-------
 prepare getemp as select * from employee where id = $1 and name = $2; |     1
 select pg_stat_monitor_reset()                                        |     1
 prepare getemp as select * from employee where id = 3 and name = bob  |     2
 select query,calls from pg_stat_monitor                               |     3
(4 rows)

postgres=# select '$' from employee;
 ?column? 
----------
(0 rows)

postgres=# select query,calls from pg_stat_monitor;
                                 query                                 | calls 
-----------------------------------------------------------------------+-------
 prepare getemp as select * from employee where id = $1 and name = $2; |     1
 select pg_stat_monitor_reset()                                        |     1
 prepare getemp as select * from employee where id = 3 and name = bob  |     2
 select query,calls from pg_stat_monitor                               |     4
 select '$' from employe                                               |     1.   <<<<<<<----- it truncate the name of table.
(5 rows)

```

MartinSun34:
Sure, I will take care of this problem and do more test

PR #72, opened 2021-02-04
PG-147 Stored Procedure Support add parentid to track caller

MartinSun34:
This patch will add a colum in pg_stat_monitor, it will record the outer layer caller's pueryid as inner layer query's parentid.

Review by @EngineeredVirus


pg_stat_monitor.h
```diff
@@ -80,6 +80,8 @@
 #define APPLICATIONNAME_LEN	100
 #define PGSM_OVER_FLOW_MAX	10
 
+/* the assumption of query max nested level */
+#define DEFAULT_MAX_NESTED_LEVEL	10
```
EngineeredVirus, line 84: Why a value of 10 and why not less or more? Realistically, I'd suggest to use max_stack_depth value.

sql/basic.sql
```diff
@@ -2,6 +2,6 @@ CREATE EXTENSION pg_stat_monitor;
 SELECT pg_stat_monitor_reset();
 select pg_sleep(.5);
 SELECT 1;
-SELECT query FROM pg_stat_monitor ORDER BY query;
+SELECT query FROM pg_stat_monitor ORDER BY query COLLATE "C";
```
EngineeredVirus, line 5: This should be part of another separate patch for regression. Also, I think that we should change the query so that its output is not collation dependent.

pg_stat_monitor.c
```diff
@@ -984,6 +1003,11 @@ static void pgss_store(uint64 queryId,
 		_snprintf(e->counters.error.sqlcode, sqlcode, sqlcode_len, SQLCODE_LEN);
 		_snprintf(e->counters.error.message, message, message_len, ERROR_MESSAGE_LEN);
 
+		if(nested_level > 0)
+			e->counters.info.parentid = nested_queryids[nested_level - 1];
+		else
+			e->counters.info.parentid = UINT64CONST(0);
```
EngineeredVirus, line 1009: I'd suggest to #define an empty parentid rather than using this const value; perhaps:

#define QUERY_PARENT_NONE UINT64CONST(0)

EngineeredVirus (changes requested):
We also must update the pgssHashKey structure to include the parentid. Also, please do update the relevant code.

ibrarahmad:
The patch does now work for me, please check that.

```
postgres=# select * from add(1,2);
 add 
-----
   3
(1 row)

postgres=# select prosrc from pg_proc where proname like 'add';
     prosrc      
-----------------
 select $1 + $2;
(1 row)


postgres=# select queryid, substr(query, 0, 30), parentid from pg_stat_monitor;
     queryid      |            substr             | parentid 
------------------+-------------------------------+----------
 D64C8DB8CD001F18 | select prosrc from pg_proc wh | 
 E12B4104FD85859A | select $1 + $2                | 
 96C96A98020966B5 | select * from add($1,$2)      | 
 3BBF759969838E17 | SELECT pg_catalog.quote_ident | 
 CC9AF910C6E0189D | select queryid, substr(query, | 
(5 rows)
```

ibrarahmad:
After fixing the issue mentioned by Hamid and identified by me. I have committed the patch.

PR #75, opened 2021-02-20
PG-183

mu-samoylov:
Move tests to separate directory and test version.


ibrarahmad:
No these are regression test suites and should not be in /t/ directory. In PostgreSQL t/ is only for tap test cases which are Perl scripts. the regression test suite should not be in the t/directory.

ibrarahmad:
For version test cases, please create separate pull requests.

PR #78, opened 2021-03-08
PMM-7569 Fix release notes.

JiriCtvrtka:


JiriCtvrtka:
[PG-112] was removed from 0.7 because column client_ip was already presented in 0.6.

PR #81, opened 2021-03-23
PG-156: Fix the placeholder replacement function for prepared statement

MartinSun34:
This commit is for placeholder replacement when set PGSM_NORMALIZED_QUERY to off. It works for both prepared statement and function.

The easist way to get external parameter is from ParamList and we can get ParamList at planner hook or Executor's QueryDesc. For a parepared statement, it won't get planned for every execution, so Executor is the only place we can get external parameters. 

So my solution is to delay the query text storage until the end of execution when we set normalized_query to off. When the query is analyzed and trying to store its query text. It will only create an entry in the query_text_hash table. The entry will be marked as 'not set'. The query text will not be stored until the query reaches the end of execution or emit an error.
This has a side effect, the 'ACTIVE' query may show '<query not set yet>'.

The reason I did't delay all queries until the end of execution is that if we set normalized to on, we still need to get normalized queries on analyzer.

Hamid and I discussed how to use get_normalized_query to get all the placeholder locations. But the problem is that it needs a query tree as input. For prepared statements, it will only be parsed and analyzed during the creation, and will not be analyzed each time it is executed. So we have to use some storage space keep the querytree information and ensure that it exists during the lifetime of pg_stat_monitor. And there seems to be no way to decrease these information. Because we don't know when prepared statements will get deallocated. Therefore, my approach is to parse the query string again and use the walker function to get all the 'ParamRef' nodes in the parsetree and get their locations. This will increase the overhead, but I think this is the only way works for prepared statement.

reviewd by @EngineeredVirus

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ PGFILEDESC = "pg_stat_monitor - execution statistics of SQL statements"
 LDFLAGS_SL += $(filter -lm, $(LIBS)) 
 
 REGRESS_OPTS = --temp-config $(top_srcdir)/contrib/pg_stat_monitor/pg_stat_monitor.conf --inputdir=regression
-REGRESS = basic version guc counters relations database top_query application_name cmd_type error state rows 
+REGRESS = basic version guc counters relations database top_query application_name cmd_type error state rows normalize
 
 # Disabled because these tests require "shared_preload_libraries=pg_stat_statements",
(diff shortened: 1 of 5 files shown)


ibrarahmad:
@darkfronza Can you review this PR

EngineeredVirus:
This PR is outdated and no longer needed now.

EngineeredVirus:
Closing this PR as it is no longer needed.

PR #84, opened 2021-07-20
PG-197: Fix conflict with pg_stat_statements

darkfronza:
If pg_stat_monitor is loaded after pg_stat_statement, then it will end
up calling standard_planner function twice in the pgss_planner_hook()
function, this will trigger an assertion failure from PostgreSQL as this
function expects an untouched Query* object, and the first call to
standard_planner() done by pg_stat_statements modifies the object.

To address the problem, we avoid calling standard_planner function twice
in pg_stat_monitor, if a previous handler is installed for the hook
planner_hook, then we assume that this previous hook has already called
standard_planner function and don't do it again.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -699,9 +699,17 @@ pgss_planner_hook(Query *parse, const char *query_string, int cursorOptions, Par
 		plan_nested_level++;
 		PG_TRY();
 		{
-    		if (planner_hook_next)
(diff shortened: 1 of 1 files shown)


ibrarahmad (changes requested):
The commit should follow the rules.

PG-197 fix conflict with pg_stat_statements 
should be.

PG-197: **F**ix conflict with pg_stat_statements **.**
<Blank Line>
Detail Commit message.

PR #126, opened 2021-11-01
PG-264 fix version

EvgeniyPatlan:


AgileStas (approved):
LGTM

LenzGr (approved):
LGTM, thank you. As this is the only change that has been applied since the version string was changed, I'm fine with keeping the same version. Otherwise I'd have suggested to bump up the version number (e.g. `1.0.0-beta-3`)

PR #182, opened 2022-02-16
PG-326: Validate histogram_min and histogram_max ranges.

darkfronza:
Added hook for checking histogram min and max values, in order to ensure
that histogram_min is always less than histogram_max.


LenzGr:
So what happens if `histogram_min` is bigger than `histogram_max`? It does not seem to emit a human-readable error in this case, or does it?

PR #183, opened 2022-02-17
PG-311: Fix user/system cpu time zero values.

darkfronza:
Some queries were taking less than one millisecond to execute, since we
store cpu time values in milliseconds in a float, a value of 0.1 (a tenth
of millisecond) cast to int64 would truncate the fractional part and
the result would be zero.

To fix the problem, the cast to (int64) when calculating the average cpu
time was removed, when updating entry statistics in pgss_update_entry()
function.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -1359,8 +1359,8 @@ pgss_update_entry(pgssEntry *entry,
             }
 		    if (sys_info)
 		    {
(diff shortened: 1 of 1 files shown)


LenzGr:
Thank you, @darkfronza - would it be possible to add a test case for this as well, e.g. verifying that these values are always non-zero?

PR #187, opened 2022-02-24
Fixing issue where makefile was not honoring PG's global makefiles when placed under the contrib folder.

EngineeredVirus:
PG-352: pg_stat_monitor: make error when placed under contrib folder

Fixing issue where makefile was not honoring PG's global makefiles when placed under the contrib folder. This fix resolves it by setting MAJORVERSION from pg_config and then setting input SQL file version accordingly.

[PG-352](https://jira.percona.com/browse/PG-352)


EngineeredVirus:
This may be tested by placing and building pg_stat_monitor under the contrib folder. Also, if placed in another location, USE_PGXS would work as expected.

PR #189, opened 2022-03-03
PG-355: Collect accumulative value of sys_time and user_time.

ibrarahmad:
Collect accumulative value of sys_time and user_time, because
separate value for each same query will override the previous value.


EngineeredVirus (approved):
LGTM

PR #190, opened 2022-03-14
PG-338: Calls count is not correct in PG-13.

ibrarahmad:
PG-338: cherry-pick patch (b6838049b6a41e5d936aad920dd252f88e3ba774) by Diego
and I did some refactoring.


EngineeredVirus (approved):
LGTM

PR #191, opened 2022-03-17
PG-338: Calls count is not correct in PG-13.

ibrarahmad:
There is a duplicate query issue by this when the query plan is enabled.
Also fixes PG-331 - Defaults values for better presentation in PMM.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -1791,7 +1791,12 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 			tmp = e->counters;
 			SpinLockRelease(&e->mutex);
 		}
-		if (!IsBucketValid(bucketid))
+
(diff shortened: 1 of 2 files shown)


pg_stat_monitor.c
```diff
@@ -1791,7 +1791,12 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 			tmp = e->counters;
 			SpinLockRelease(&e->mutex);
 		}
-		if (!IsBucketValid(bucketid))
+
+        /* In case plan is enable, there is no need to show 0 planid query */
```
LenzGr, line 1795: Minor nit:
```suggestion
        /* In case that query plan is enabled, there is no need to show 0 planid query */
```

EngineeredVirus:
LGTM

EngineeredVirus:
LGTM

EngineeredVirus (commented):
LGTM

PR #201, opened 2022-04-11
PG-369: Fix wal_bytes values on PG <= 12.

darkfronza:
Similar to pg_stat_statements, pg_stat_monitor tracks wal data metrics
since PostgreSQL 13, the problem was that for PostgreSQL versions 11 and
12 we left the WalUsage variable declared in the stack unitialized,
thus leading to garbage values.

Fixed the problem by ignoring the WalUsage variable value for PG <= 12.


ibrarahmad:
LGTM

PR #206, opened 2022-04-14
PG-358 Documented block and WAL data display

nastena1606:
Due to different data processing by both extensions, these stats appear as inconsistent.

modified:   README.md
	modified:   docs/COMPARISON.md


capri1989 (approved):
LGTM

PR #214, opened 2022-05-03
PG-365: bump version in META.json file to the latest release

capri1989:
The META.json file had still an old release reference. Fixed that by changing it to the newest release. 


LenzGr (commented):
Thanks for taking care of this! Would you mind adding the `homepage` resource as suggested in PG-365 as well? See https://github.com/reorg/pg_repack/blob/master/META.json for an example. As an URL for this resource I would suggest using https://percona.github.io/pg_stat_monitor/ - thank you!

capri1989:
> Thanks for taking care of this! Would you mind adding the `homepage` resource as suggested in PG-365 as well? See https://github.com/reorg/pg_repack/blob/master/META.json for an example. As an URL for this resource I would suggest using https://percona.github.io/pg_stat_monitor/ - thank you!

Thanks for the input. Fixed it.

Naeem-Akhter (approved):

EngineeredVirus (approved):
LGTM:

The description may need an update.

PR #215, opened 2022-05-03
PG-365:  change version in META.json file to dev

capri1989:
The META.json file had still an old release reference. Fixed that by changing it to "dev" for the main branch.

Added the "homepage" reference as requested in the PG-365 ticket.


Naeem-Akhter (approved):
LGTM

PR #217, opened 2022-05-03
PG-407 Added a 404 page to docs

nastena1606:
new file:   docs/overrides/404.html
	modified:   mkdocs.yml

### DIFF
--- a/docs/overrides/404.html
+++ b/docs/overrides/404.html
@@ -1,0 +1,9 @@
+{#-
+  This file was automatically generated - do not edit
+-#}
+{% extends "main.html" %}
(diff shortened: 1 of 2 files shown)


capri1989:
Hi @nastena1606, we could cherry-pick your fix from the REL1_0_STABLE branch, so you don't need to create additional PRs. This way we also keep the original git commit id.

nastena1606:
That's what I did...

вт, 3 мая 2022 г., 17:23 Kai Wagner ***@***.***>:

capri1989 (approved):
LGTM

PR #218, opened 2022-05-03
PG-269: Updated install from source chapter

capri1989:
Updated the "Building from source" chapter. Added the postgresql-devel package for yum and postgresql-server-dev-all for apt variants and also removed the pg_config. 


nastena1606 (approved):

Naeem-Akhter (approved):
LGTM

PR #220, opened 2022-05-04
PG-383: Fix coverall.io badge on main branch in PGSM repo.

Naeem-Akhter:
This will make sure that coverall.io stats are reflected as main branch.


EngineeredVirus (approved):
LGTM

PR #232, opened 2022-05-22
PG-417: Replace 'CREATE OR REPLACE' with 'CREATE' for extension.

ibrarahmad:
While analyzing the pg_stat_monitor installation scripts I found several
vulnerabilities. pg_stat_monitor uses CREATE OR REPLACE to install its
functions which is a security hazard. An attacker can precreate the functions
have a superuser install the extension and after installation the attacker
can switch out the function with a malicious version since he would still
be the owner of the function. Instead of CREATE OR REPLACE the installation
script should use plain CREATE to prevent this attack.

For reference
https://www.postgresql.org/docs/current/extend-extensions.html#EXTEND-EXTENSIONS-SECURITY
https://github.com/timescale/pgspot


EngineeredVirus (approved):
LGTM

capri1989 (commented):
LGTM

capri1989 (approved):

PR #234, opened 2022-05-23
DISTPG-427: replaced return with goto exit to not break other extensions

capri1989:


Naeem-Akhter (approved):
Looks good to me.

PR #235, opened 2022-05-23
PG-424: bump version to 1.0.1

capri1989:


capri1989:
> We also need to update the following changes as part of this version bump:
> 
> * Meta.json file version is not updated.
> * Change log for the RPMs and Debian packages

Addressed Meta.json file already and started a discussion about the changelog files. Will get back to you on this

EvgeniyPatlan (approved):

EngineeredVirus (changes requested):
Let's maintain the changelog entries for the older releases, rather than removing those entries.

EngineeredVirus (approved):
LGTM

PR #236, opened 2022-05-24
PG-382: Adjust the maximum value for histogram buckets.

ibrarahmad:
There was no maximum limit set for the number of maximum histograms
bucket, which can lead to a crash in case of higher value.


capri1989:
Please squash your commits into a single one.

MichaelDBA:
How is this getting missed?

capri1989:
> How is this getting missed?

Hi @MichaelDBA,

what do you mean exactly?

MichaelDBA:
@ibrarahmad commented a few hours ago that "There was no maximum limit set for the number of maximum histograms
bucket, which can lead to a crash in case of higher value."  So I was wondering why this was being said now.

capri1989:
> @ibrarahmad commented a few hours ago that "There was no maximum limit set for the number of maximum histograms bucket, which can lead to a crash in case of higher value." So I was wondering why this was being said now.

That was the explanation "why" this PR was created. So the reason you've hit this bug was exactly because of the missing max number.

If I didn't understand your comment, just let me know, happy to discuss further to make sure everything is addressed before this gets merged.

ibrarahmad:
I have updated the TAP test cases and SQUASH the commits.

capri1989 (approved):
LGTM now

Naeem-Akhter (approved):
Verified that this PR fixes the issue reported in ticket PG-382. 

PR was testing on both, PGDG and PPG distributions of Postgres versions 11~14.

PR #246, opened 2022-05-30
PG-432 Fixed branch name of a  GH action

nastena1606:
modified:   .github/workflows/doc-build.yml

### DIFF
--- a/.github/workflows/doc-build.yml
+++ b/.github/workflows/doc-build.yml
@@ -2,7 +2,7 @@ name: Build pg_stat_monitor docs
 on:
   push:
     branches: 
-      - REL1_0_STABLE
+      - REL_1_STABLE
       
       
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/doc-build.yml
```diff
@@ -2,7 +2,7 @@ name: Build pg_stat_monitor docs
 on:
   push:
     branches: 
-      - REL1_0_STABLE
+      - REL1_STABLE
```
capri1989, line 5: The branch name isn't correct. The branch is called "REL_1_STABLE"

capri1989 (changes requested):

.github/workflows/doc-build.yml
nastena1606: Fixed

capri1989 (approved):
LGTM

PR #247, opened 2022-05-30
Pmm 9951 squash

JiriCtvrtka:


capri1989:
@JiriCtvrtka What do you want to achieve with this PR? If this is about squashing, this can be done in the original PR. https://github.com/percona/pg_stat_monitor/pull/245

If you don't know how to do that, here's a quick guide:

- "git rebase -i HEAD~11" (looking at pull/245 as there are 11 commits in it)
    - You can also directly go with the SHA if you want within the git rebase command.
-  Afterwards just push again into your branch - most of the time this should be "git push origin -f"

PR #248, opened 2022-05-31
PG-430 Update rpm changelog automatically during build

adivinho:


EngineeredVirus (changes requested):
This is a good start. However, my concern would be that without proper testing, there is a likelihood that the changelog updates would go untested, as the changelog must also include any packaging related changes; say, a path change.

How would those changes be accommodated with this PR and how would those changes be tested?

EvgeniyPatlan (approved):

puneet0191:
@adivinho please update your branch with latest changes from upstream

capri1989:
Those changes do land in the spec file for the packages directly but do not make it back into the git repo demo spec file, correct?

capri1989:
Duo to our switch to the OneFlow development model, we now have only one branch and as such the "main" branch. Please recreate again against the main branch instead.

PR #253, opened 2022-06-07
PG-362: pgsm_normalized_query needs to be 0 by default.

ibrarahmad:
pg_stat_monitor.pgsm_normalized_query needs to be 0 by default because
people will get Query examples. This is one of the main user-faced
advantages of pgsm over pgss (pg_stat_statements).


ibrarahmad:
Don't worry about the regression, because we need to adjust the regression. Regression has many issues which are based on random values and values keep changing. We need to stabilize the regression on separate issues.

EngineeredVirus (approved):
LGTM

PR #254, opened 2022-06-07
PG-300: Fix potential duplicate queryid issue.

ibrarahmad:
There is some potential problem where there is a chance
that we got a duplicate queryid. This patch will eliminate
that problem.


EngineeredVirus (changes requested):
Can you please provide a few more details as how this code change actually fixes the duplicate queryid issue?

PR #255, opened 2022-06-07
PG-354: Use pg_settings view to generate pg_stat_monitor_settings.

ibrarahmad:
There is no need to add code to create pg_stat_monitor_setting view.
PostgreSQL already have pg_setting view, which can be used to
gererate pg_stat_monitor_view.


EngineeredVirus (changes requested):
This seems like a breaking change for PMM. It should be targeted for 2.0.

Naeem-Akhter (commented):
Seems, tests output files needs to be updated for TAP test cases.

ibrarahmad:
This PR is no longer necessary, it is for Version 2.0.0

PR #263, opened 2022-06-14
PG-440: Regression Cleanup.

ibrarahmad:
This PR contains two parts; the first is for the normal regression and the second one is the TAB test case.

1-  In first commit, we removed unnecessary files which no longer in use, along
    with combining similar test cases in a single file to avoid duplication.

    
 2 -  This commit moved all the old TAB test cases into the old_tests directory from
    the original directory, because of too much duplication of code and redundant
    test cases. A new Perl Module t/pgsm.pm is introduced to contain all the regression
    TAB tests generic functions. Two new test cases "t/0001_settings_pgsm_track_planning.pl"
    and "t/0002_settings_pgsm_enable_query_plan.pl" added to the regression and these are
    using the "pgsm.pm" module.  These will avoid the code duplications and each test case
    is doing the desired testing instead of testing everything in one test case. All the
    old test cases need to follow these test cases and moved back from old_test to the t/ folder.


Naeem-Akhter:
As per discussions, instead of merging this PR into REL_1_STABLE, it needs to be merged against main branch. That will be taken care by me and following are action items on my part: 

1) Once all other merges are done into main branch, then test changes proposed in this PR against main branch.
2) Fix any other regressions that might come up due to the breaking changes in main branch. 
3) Create PR against main branch with all changes.

Naeem-Akhter:
As per discussions, instead of merging this PR into REL_1_STABLE, it needs to be merged against main branch. That will be taken care by me and following are action items on my part:

1. Once all other merges are done into main branch, then test changes proposed in this PR against main branch.
2. Fix any other regressions that might come up due to the breaking changes in main branch.
3. Create PR against main branch with all changes.

PR #266, opened 2022-06-20
[PG-159] pg_stat_monitor: Bucket start time should be aligned with fi…

EngineeredVirus:
The buckets are now created with the start time a modulus of the bucket time size.
So if we have a 10 second bucket, the start times would reflect that:
- Bucket1: 00:00:00
- Bucket2: 00:00:10
- Bucket3: 00:00:20
...

Previously, the start time of the bucket was aligned with the first query that
arrives in that bucket. However, now the behaviour is changed. So, even if the
first query for bucket 2 arrives at 00:00:13, the start time would still be set
to 00:00:10.

This change now makes the bucketing separated out by fixed time windows so that
external applications can easily consume that data and chart it.

Also, as part of this change, locking of pgss is updated now and extended
to last the bucket related changes.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -2085,87 +2085,79 @@ static uint64
 get_next_wbucket(pgssSharedState *pgss)
 {
 	struct timeval	tv;
-	uint64			current_usec;
-	uint64			current_bucket_usec;
+	uint64			current_sec;
+	uint64			current_bucket_sec;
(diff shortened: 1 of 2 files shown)


ibrarahmad (approved):
This PR looks good to me. In my opinion, this behaviour should be documented. @EngineeredVirus Can you please create a document Jira?

PR #268, opened 2022-06-21
PG-453: Normalize query does not work with INSERT statements.        …

ibrarahmad:
…          

The commit fixes the issue similar to pg_stat_statements. Jumble query skips in    
case of INSERT statements to avoid the duplicate queryid, but in another      
commit we already solved that problem with another way.


EngineeredVirus (approved):
LGTM

PR #270, opened 2022-06-23
PG-452: Remove redundant workflow from PGSM GH main branch.

Naeem-Akhter:
Removed following workflows from main branch. These workflows are already been covered and replaced by other distribution (PGDG & PP) specific workflows. 

postgresql-14-package.yml
postgresql-13-package.yml
postgresql-12-package.yml
postgresql-11-package.yml


Naeem-Akhter:
Regression jobs will fail due to pending regression issues after branches merge. So please ignore these.

capri1989 (approved):
LGTM

PR #273, opened 2022-06-28
PG-457: added vim code annotation config example

capri1989:
I've added the .vimrc config example as a spoiler to not mess up the documentation and list. 


nastena1606 (approved):
LGTM

PR #277, opened 2022-06-30
Regressions fix post branches merge

Naeem-Akhter:
Regression was failing in main branch after merging REL_1_STABLE into it. This required changes to a number of testcases (both sql and tap, with output changes).


ibrarahmad (changes requested):
@Naeem-Akhter I know you are fixing the regression, but this is quite a lot of changes. I need you to write a detailed commit message about what you have done and why, because only saying fixing regression is not enough. I usually like a separate fix for every change event that is regression. If that is not possible then write what major changes you have done here. Please squash your commits too.

Naeem-Akhter:
Closing this PR, as created another PR with requested changes under https://github.com/percona/pg_stat_monitor/pull/278.

PR #284, opened 2022-08-01
PG-464: Update regression status badges links.

Naeem-Akhter:
Updated status badges links in README that were not carrying the correct links, workflows (Actions) for older links have been removed from repo as those have been replaced by distribution specific (PGDG and PPG) workflows.


Naeem-Akhter:
This is what README would like after badges updates: https://github.com/Naeem-Akhter/pg_stat_monitor/tree/PG-464.

EngineeredVirus (approved):
LGTM

PR #285, opened 2022-08-02
PG-376: Enable TAP testing on PG 12

EngineeredVirus:
This commit does not make any changes to the makefile. It simply ensures that
the test cases work with PG12. Necessary skip statements are added. Makefile
changes are to be done as part refactoring the Makefile.


Naeem-Akhter (approved):
Looks good to me.

PR #286, opened 2022-08-02
PG-352: pg_stat_monitor: make error when placed under contrib folder

EngineeredVirus:
The Makefile is refactored and unnecessary conditional statements are removed
in favor of handling version separately.

Also, the clean target is updated to remove generated files including results
folder and the ${DATA} file.


Naeem-Akhter (approved):
LGTM.

PR #287, opened 2022-08-02
PG-462: adding PG15 build to our workflows

EngineeredVirus:
Adding SQL file for PG15. Initial version is a replica of PG14 SQL file.
Changes will be required in this file to fully support newly introduced columns
in PG15.

This currently fixes compilation and make install targets.

Also, updated the first line comments where the SQL version was mentioend as
1.1. The version remains at 1.0 for the time being.


Naeem-Akhter (approved):
Looks good to me.

PR #289, opened 2022-08-04
PG-476: pg_stat_monitor causing errors : could not read file /tmp/pg_…

EngineeredVirus:
…stat_monitor_query: No such file or directory

This only fixes the basic the path issue. It moves the query overflow file
from /tmp to PGSTAT folder data directory.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -22,6 +22,7 @@
 #ifdef BENCHMARK
 #include <time.h>				/* clock() */
 #endif
+#include "pgstat.h"
 #include "commands/explain.h"
 #include "pg_stat_monitor.h"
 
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -22,14 +22,15 @@
...
+#include "pgstat.h"
 #include "commands/explain.h"
 #include "pg_stat_monitor.h"
 
 PG_MODULE_MAGIC;
 
 #define BUILD_VERSION                   "1.1.0-dev"
 #define PG_STAT_STATEMENTS_COLS         53	/* maximum of above */
-#define PGSM_TEXT_FILE                  "/tmp/pg_stat_monitor_query"
+#define PGSM_TEXT_FILE PGSTAT_STAT_PERMANENT_DIRECTORY "pg_stat_monitor_query"
```
ibrarahmad, line 33: 1 - Please check that if pg_stat_statment is never installed on that system, will that work?
2 - It will create confusion to share the pg_stat_stetment directory with pg_stat_monitor.

ibrarahmad (changes requested):

pg_stat_monitor.c
EngineeredVirus: This define is part of pg_stat, not pg_stat_statements. So we are good with it. I've verified that this should work for all supported versions of PG11 onwards.

ibrarahmad (approved):

PR #292, opened 2022-08-15
PG-477: Install perl modules in PGSM-PG12 Build workflow.

Naeem-Akhter:
Perl tap tests dependencies were missing, those are being installed now. Furthermore, workflow required some changes to enable logs to be available in case of tap tests failure(s).


EngineeredVirus (approved):
LGTM

PR #293, opened 2022-08-15
PG-464: Update regression status badges links.

Naeem-Akhter:
As per discussion in the team meeting, removed status badges for Pg-11~13 jobs from README. We should only be displaying the status badges for the latest PG version (GA) and code coverage.


EngineeredVirus (approved):
LGTM

PR #295, opened 2022-08-22
PG-475: Inconsistent behaviour of PGSM

EngineeredVirus:
Reverting the bucket locking mechanism to previous behavior. This has
a lot of room for improvement that needs to be part of a major refactoring
in the 2.x release.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -2151,7 +2151,7 @@ get_next_wbucket(pgssSharedState *pgss)
 	uint64		new_bucket_id;
 	uint64		prev_bucket_id;
 	struct tm  *lt;
-	char		file_name[1024];
+	bool		update_bucket = false;
 
 	gettimeofday(&tv, NULL);
 	current_sec = (TimestampTz) tv.tv_sec - ((POSTGRES_EPOCH_JDATE - UNIX_EPOCH_JDATE) * SECS_PER_DAY);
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -2145,80 +2145,85 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
...
-		}
-	}
-
-	snprintf(pgss->bucket_start_time[new_bucket_id], sizeof(pgss->bucket_start_time[new_bucket_id]),
-			 "%04d-%02d-%02d %02d:%02d:%02d", lt->tm_year + 1900, lt->tm_mon + 1, lt->tm_mday, lt->tm_hour, lt->tm_min, lt->tm_sec);
-
-	LWLockRelease(pgss->lock);
-
-	return new_bucket_id;
+    struct timeval  tv;
```
ibrarahmad, line 2148: Why do we have whitespace changes? Even if the previous is wrong, that should not be here in a fix.

ibrarahmad (changes requested):
The patch should contain the changes required; no code cleanup and whitespace changes are not allowed in the bug fix commit.

pg_stat_monitor.c
EngineeredVirus: Updated the whitespaces to match the current whitespace settings. Previous patch had the same whitespace as the original code.

ibrarahmad (approved):
LGTM

PR #298, opened 2022-08-29
PG-501: Missing Buckets and incorrect calls count.

codeforall:
prev_bucket_sec holds the actual time at which the previous bucket was created
and it is used to compute if the previous bucket time has elapsed and when is
the time to create a new one. But since the bucket start time is rounded down
to logical time window start, that makes the prev_bucket_sec and bucket start
time out of sync with each other, and depending on the query arrival time there
is a high probability that a bucket gets missed especially when the last bucket
was created around the end of the bucket time window.

Solution is to keep the prev_bucket_sec and bucket start time in-sync.

Moreover, we are using the unint64 for storing the prev_bucket_sec which is kind
of an overkill and a simple uint should be good enough for the purpose. But that
change can be taken up as part of the create-bucket function refactoring task.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -2146,7 +2146,6 @@ static uint64
 get_next_wbucket(pgssSharedState *pgss)
 {
 	struct timeval tv;
-	uint64		current_sec;
 	uint64		current_bucket_sec;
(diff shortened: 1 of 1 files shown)


EngineeredVirus (approved):
LGTM

PR #304, opened 2022-09-15
PG-520 pg_stat_monitor does not work with PG15

codeforall:
PG 15 requires additional shared memory and LWLocks requests to be made from the newly introduced shmem_request_hook and disallows the requests initiated from outside the hook.
The commit makes moves the additional shared memory and LWLocks requests from _PG_init to shmem_request_hook for PG15


ibrarahmad (approved):
LGTM

PR #307, opened 2022-09-26
PG-462: adding PG15 build to our workflows

capri1989:


Naeem-Akhter (approved):
Looks good to be. PG-15 build and regression passed.

PR #309, opened 2022-10-24
PG-174: Code cleanup.

ibrarahmad:
pg_stat_monitor is a bit longer; therefore, it requires some code cleanup. Therefore I decided to turn these tasks into multiple commits and PR to avoid various changes in one PR. This will ease the review and Q/A process. In this commit, I have done these tasks.

 1 - Delete all the SQL.in files because these version-dependent files
 are becoming significant in quantity. Now added a single SQL file for which
 contains the dynamic SQL based on the PostgreSQL Version.

 2 - New SQL files (pg_stat_monitor--2.0.sql) added for pg_stat_monitor version 2.

 3 - A new SQL file (pg_stat_monitor--1.0--2.0.sql) is created, which will be
 used to upgrade from version 1.0 to 2.0. Currently, this file is empty. But
 whenever we add some API changes into 2.0, we need to update that file too.

 4 - The control file (pg_stat_monitor.control) is updated for version 2.0.
 This change will make the CREATE EXTENSION default to pg_stat_monitor version 2.0


EngineeredVirus (approved):
LGTM

PR #310, opened 2022-10-24
PG-174: Code cleanup.

ibrarahmad:
pg_stat_monitor is a bit longer; therefore, it requires some code cleanup. Therefore I decided to turn these tasks into multiple commits and PR to avoid various changes in one PR. This will ease the review and Q/A process. In this commit, I have done these tasks.

1 - Fixing the compilation issue caused by the previous commit, where we removed the benchmarking code. It was causing problems for PostgreSQL-12.


EngineeredVirus (approved):
LGTM

PR #312, opened 2022-11-15
PG-320: Removing the query state code from the view.

ibrarahmad:
The query status monitoring code was used to track the current query state, for example,      parsing, executing and finishing. After careful review, we have figured out that   it does not make sense while a lot of time same query is running. Therefore it   is also consuming resources. This commit will remove that feature. The upgrade SQL from 1.0 - 2.0 is also updated.


Naeem-Akhter (approved):
LGTM !

PR #314, opened 2022-11-15
   PG-518: Drop the internal function permission from PUBLIC.

ibrarahmad:
PG-518: Drop the internal function permission from PUBLIC.
    
    It will be a security problem to provide the internal function access to PUBLIC.
    This commit will revoke all permission of internal functions from PUBLIC.


Naeem-Akhter (approved):
LGTM !

PR #315, opened 2022-11-15
PG-306: The bucket start time should be timestamp instead of TEXT.

ibrarahmad:
PG-306: The bucket start time should be timestamp instead of TEXT.


Naeem-Akhter (approved):
LGTM !

PR #318, opened 2022-11-16
PG-544: Regression cleanup.

ibrarahmad:


Naeem-Akhter (approved):
I have tested this PR against all PG 11~15. 
PG 11, 12, 13 are clear now. Only problem is PG 14 & 15, where 3 SQL test cases are failing. I will verify those tomorrow and push a fix. TAP testcases are also clear on all PG versions (on my laptop VMs) for branch used, PG544.

PR #319, opened 2022-11-17
PG-554: Remove redundant files and fix regression.

Naeem-Akhter:
Removed old files with same name and add these files to fix sql regression on PG 14 & 15.

1- regression/expected/error_1.out
2- regression/expected/error_insert_1.out
3- regression/expected/top_query_1.out


EngineeredVirus (commented):
Are any files removed or intended to be removed?

Also, this seems like a behaviour change. Are these expected changes in the output file match the expected outputs for other version?

ibrarahmad (approved):

EngineeredVirus (approved):
LGTM

PR #320, opened 2022-11-18
PG-555 :Infrastructure to allow multiple SQL APIs

codeforall:
Creating the infrastructure that'll allow using newer versions of the loadable module with old SQL declarations.
Also updating the build version to 2.0.0-dev


ibrarahmad (approved):

EngineeredVirus (approved):
LGTM

PR #321, opened 2022-11-19
PG-310: Bucket is “Done” vs still being current/last.

ibrarahmad:
A new column is added to mention that bucket is active or done. there is     some timing based adjustment was required with that too.


EngineeredVirus (approved):
LGTM.

It's missing a test case. Also, for long-running queries (under very rare cases where a query runs for more than a bucket, this might be problematic). However, this seems like a good fix.

PR #325, opened 2022-11-30
PG-553: Add a test case to verify columns names in PGSM.

Naeem-Akhter:


codeforall:
I have a couple of suggestions.

PR #326, opened 2022-11-30
PG-435: Adding new counters that can be tracked in PG15

codeforall:
In line with pg_stat_statments for PG15, This commit adds eight new cumulative counters for jit operations making it easier to diagnose how JIT is used in an installation. And two new columns, temp_blk_read_time, and temp_blk_write_time respectively show the time spent reading and writing temporary file blocks on disk.

### DIFF
--- a/t/018_column_names.pl
+++ b/t/018_column_names.pl
@@ -1,0 +1,100 @@
+#!/usr/bin/perl
+
+use strict;
+use warnings;
+use File::Basename;
+use File::Compare;
+use File::Copy;
+use String::Util qw(trim);
+use Test::More;
+use lib 't';
+use pgsm;
+
+# Get filename and create out file name and dirs where requried
+PGSM::setup_files_dir(basename($0));
+
+# Create new PostgreSQL node and do initdb
+my $node = PGSM->pgsm_init_pg();
+my $pgdata = $node->data_dir;
+
+# Update postgresql.conf to include/load pg_stat_monitor library
+open my $conf, '>>', "$pgdata/postgresql.conf";
+print $conf "shared_preload_libraries = 'pg_stat_monitor'\n";
+close $conf;
+
+# Dictionary for expected PGSM columns names on different PG server versions
+my %pg_versions_pgsm_columns = ( 15 => "application_name,blk_read_time," .
+    "blk_write_time,bucket,bucket_done,bucket_start_time,calls," .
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -1671,7 +1709,7 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 	if (get_call_result_type(fcinfo, NULL, &tupdesc) != TYPEFUNC_COMPOSITE)
 		elog(ERROR, "pg_stat_monitor: return type must be a row type");
 
-	if (tupdesc->natts != 52)
+	if ( tupdesc->natts != expected_out_args )
```
EngineeredVirus, line 1712: I think this check should be replaced by an assertion. I don't see this every happening in production.

pg_stat_monitor.c
```diff
@@ -3343,8 +3400,8 @@ pg_stat_monitor_settings(PG_FUNCTION_ARGS)
 		}
 		else
 		{
-			values[j++] = Int64GetDatumFast(get_conf(i)->guc_min);
-			values[j++] = Int64GetDatumFast(get_conf(i)->guc_max);
+			values[j++] = Int32GetDatum(get_conf(i)->guc_min);
+			values[j++] = Int32GetDatum(get_conf(i)->guc_max);
```
EngineeredVirus, line 3404: Why is this change needed?

pg_stat_monitor.c
```diff
@@ -1817,7 +1855,7 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 		}
 
 		/* state at column number 8 */
-		values[i++] = Int64GetDatumFast(tmp.state);
+		values[i++] = Int64GetDatumFast((int64)tmp.state);
```
EngineeredVirus, line 1858: Why the explicit cast?

pg_stat_monitor.c
codeforall: First, it was wrong to use 64Get* macro on 32-bit variable, Second PostgreSQL code [commit](https://github.com/postgres/postgres/commit/c8b2ef05f481ef06326d7b9f3eb14b303f215c7e) adds a static assertion to GetDatum* macros to ensure type correctness, so this was always going to throw a compiler error in the future.

pg_stat_monitor.c
codeforall: Agreed

PR #330, opened 2022-12-07
PG-316: Convert blocks timings to average per buckets.

ibrarahmad:


codeforall (commented):
I doubt the formula used to calculate the average is correct. I believe the correct and easiest place to do that would be at the time of constructing the output tuple (in pg_stat_monitor_internal function).

codeforall (commented):
The average is not calculated for temp_blk_write_time. and temp_blk_read_time. Is it a requirement?

PR #334, opened 2022-12-13
PG-354: pg_stat_monitor: Remove pg_stat_monitor_settings view

EngineeredVirus:
Removing the view for 2.0. Updating the required SQL files to manage the upgrade. Downgrade from 2.x to 1.x is not supported.

Also part of this fix is the SQL regression. This does not update the tap test cases.


ibrarahmad (commented):
LGTM, but there are test cases that need to be updated. @hamid create new Jira for regression cleanup or managed that here

ibrarahmad (approved):
LGTM

PR #336, opened 2022-12-20
PG-488: pg_stat_monitor: Overflow management.

codeforall:
Reimplement the storage mechanism of buckets and query texts using Dynamic shared memory. Since the dynamic shared memory can grow into a swap area, so we get the overflow out of the box.

oreover the new design saves the query pointer inside the bucket and eventually, the query text gets evicted with the bucket recycle.

Finally, the dynamic shared memory hash has a built-in locking mechanism so we can revisit the whole locking in pg_stat_monitor has potential for lots of performance improvements

### DIFF
--- a/hash_query.c
+++ b/hash_query.c
@@ -16,58 +16,92 @@
  */
 #include "postgres.h"
 #include "nodes/pg_list.h"
-
 #include "pg_stat_monitor.h"
 
+static pgsmLocalState	pgsmStateLocal;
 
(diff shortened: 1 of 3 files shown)


codeforall:
This fails one test case that is because of a semicolon appearing at the end of the parent query text. That IMHO is not an issue and once the code review for this passes I will submit an expected output change for that

ibrarahmad (approved):
LGTM.

PR #337, opened 2022-12-20
PG-172: Code cleanup and pgindent.

ibrarahmad:


EngineeredVirus:
Needs to be rerun just before we tag the branch for the release.

PR #339, opened 2022-12-22
PG-354: Update expected output file for functions test case.

Naeem-Akhter:
Due to removal pg_stat_monitor_settings view, functions sql test case was failing. Fixed that.


ibrarahmad (commented):
LGTM

ibrarahmad (approved):

PR #340, opened 2022-12-27
PG-563: Update TAP testcases and output due to changes by DEV.

Naeem-Akhter:
As part of this PR, also updated regression test cases that are related to following JIRA issues as well.

PG-354	pg_stat_monitor: Remove pg_stat_monitor_settings view Now we not using pg_stat_monitor_settings view, due to this change majority of TAP test case required output changes.

PG-558: Create test case to verify the function names and count in PGSM. Added additional output file for SQL test case.

PG-554:	Remove redundant expected output files from regression. Removed unnecessary output files in TAP testcases where these were not needed.

Also made case for SQL keywords consistent across the sql and TAP test cases where possible and needed.


ibrarahmad (approved):
LGTM

PR #341, opened 2022-12-28
PG-545: pg_stat_monitor: Same query text should generate same queryid

EngineeredVirus:
Regardless of the database or the user, the same query will yield the same query ID. As part of this, a new column, 'pgsm_query_id', is added.

* pgsm_query_id: pgsm_query_id has the same data type of int8 as the queryid column. If the incoming SQL command includes any constants, it internally normalizes the query to remove those constant values with placeholders. Otherwise, it uses the query directly to generate the query hash.

Since we no longer depend on the server's parse tree mechanism, we can generate the same hash for the same query text for all server versions.

Also, it is important to note that the hash being calculated is a database, schema and user independent. So same query text in different databases will generate the same hash.

This column is not part of the key; rather, for observability purposes only.

* Regression SQL test case pgsm_query_id.sql is added to the SQL regression.


pg_stat_monitor--2.0.sql
```diff
@@ -91,6 +91,7 @@ CREATE FUNCTION pg_stat_monitor_internal(
     OUT planid              text,
     OUT query               text,
     OUT query_plan          text,
+    OUT pgsm_query_id       int8,
```
ibrarahmad, line 94: We need to have that type BigInt because all int variables are in BigINT in pg_stat_statments.

pg_stat_monitor--2.0.sql
```diff
@@ -170,6 +171,7 @@ CREATE VIEW pg_stat_monitor AS SELECT
     userid::regrole,
     datname,
 	'0.0.0.0'::inet + client_ip AS client_ip,
+	pgsm_query_id,
```
ibrarahmad, line 174: TAB Issue.

pg_stat_monitor.c
```diff
@@ -1388,13 +1386,15 @@ pgss_store(uint64 queryid,
...
 	int			app_name_len = 0;
 	bool		reset = false;
+	uint64		pgsm_query_id = 0;
 	uint64		bucketid;
 	uint64		prev_bucket_id;
 	uint64		userid;
 	uint64		planid;
 	uint64		appid = 0;
-	char		comments[512] = "";
+	int			norm_query_len = 0;
```
ibrarahmad, line 1395: TAB Issue

pg_stat_monitor.c
```diff
@@ -1506,14 +1506,31 @@ pgss_store(uint64 queryid,
 		 * in the interval where we don't hold the lock below.  That case is
 		 * handled by entry_alloc.
 		 */
-		if (jstate && PGSM_NORMALIZED_QUERY)
+		if (jstate && jstate->clocations_count > 0)
```
ibrarahmad, line 1509: I doubt that is correct because it should be run in case of Normalised query.

pg_stat_monitor.c
EngineeredVirus: Deliberately ignored it as we need to run pgindent on the whole code.

pg_stat_monitor--2.0.sql
EngineeredVirus: Deliberately ignored it as we need to run pgindent on the whole code.

pg_stat_monitor--2.0.sql
EngineeredVirus: I kept the data type consistent with our queryid data type. Need to review and change all data types. That should be part of a separate exercise.

pg_stat_monitor.c
EngineeredVirus: PGSM_NORMALIZED_QUERY case is handled in the if block. We are using the norm_query pointer to identify if we need to store normalized query or not. This updated condition avoids unnecessary normalization attempt when there is no constant value in the query.

EngineeredVirus: I've validated that normalized query tap test case passes. So there shouldn't be any issue here.

ibrarahmad (approved):
LGTM

PR #342, opened 2022-12-28
PG-488: pg_stat_monitor: Overflow management.

codeforall:
Reimplement the storage mechanism of buckets (for PG-15 onward) and query texts using Dynamic shared memory. Since the dynamic shared memory can grow into a swap area, so we get the overflow out of the box.

As PostgreSQL versions prior to V15 does not support sequence scan on dynamic shared memory hashes, so older versions has to live with the classic shared memory hash for storing the buckets.

Another noteworthy change with the new design is: it saves the query pointer inside the bucket, and eventually, the query text gets evicted with the bucket recycle.

Finally, the dynamic shared memory hash has a built-in locking mechanism, so we can revisit the whole locking in pg_stat_monitor has the potential for lots of performance improvements.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -32,7 +32,6 @@
      PGSM_V2_0
  } pgsmVersion;
 
-
 PG_MODULE_MAGIC;
 
 #define BUILD_VERSION                   "2.0.0-dev"
@@ -78,6 +77,7 @@ static int	plan_nested_level = 0;
 
 /* The array to store outer layer query id*/
 uint64	   *nested_queryids;
+char	   **nested_query_txts;
 
 /* Regex object used to extract query comments. */
 static regex_t preg_query_comments;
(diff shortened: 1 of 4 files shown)


pg_stat_monitor.c
```diff
@@ -1508,73 +1531,43 @@ pgss_store(uint64 queryid,
...
-			}
-
-			/*
-			 * Save current query buffer length, if we fail to add a new new
-			 * entry to the hash table then we must restore the original
-			 * length.
-			 */
-			memcpy(&prev_qbuf_len, pgss_qbuf, sizeof(prev_qbuf_len));
-		}
+		/* Save the query text in raw dsa area */
```
ibrarahmad, line 1548: WARNING, don't create variables in the middle of the code.

ibrarahmad (changes requested):
Please fix all the TAB test cases too.

pg_stat_monitor.c
codeforall: I have pushed the fix

EngineeredVirus (approved):
LGTM

PR #344, opened 2023-01-03
PG-576 - Segmentation fault caused by pg_stat_monitor unique queryid creation mechanism.

EngineeredVirus:
This fix resolves the issue with incorrect query length in case of a normalized query when the query length exceeds PGSM_QUERY_MAX_LEN.


Naeem-Akhter (approved):
Regression is clear now. LGTM!

ibrarahmad (approved):
LGTM

PR #345, opened 2023-01-03
PG-570: Fix counters test case.

Naeem-Akhter:
Updated the counters test case and expected output, and also removed the unneeded output files.


ibrarahmad (approved):
LGTM

PR #346, opened 2023-01-04
PG-310: pg_stat_monitor: Bucket is Done vs still being current/last

Naeem-Akhter:
Added a TAP test case to verify the behaviour of the new 'bucket_done' column.


ibrarahmad (approved):
LGTM

PR #347, opened 2023-01-05
PG-575: Enable installcheck-world on PG 14 & 15.

Naeem-Akhter:
As we are using compute_query_id on pg14 onwards for PGSM and it causes the server installcheck-world to fail (same behaviour with PGSS). To test installcheck-world on pg14 onwards we need to disable compute_query_id and run server installcheck-world. But for PGSM regression we will still have compute_query_id on.


ibrarahmad (approved):
LGTM

PR #348, opened 2023-01-05
PG-562: Added first and last buckets to histogram for better data distribution.

ibrarahmad:
This commit adds two additional buckets to the histogram: a first bucket that starts from 0 and ends at min_time, and a last bucket that starts from max_time and ends at infinity. These buckets are meant to provide context for the other buckets in the histogram, which are based on the user-specified number of buckets.

For example, if the user specifies three buckets and min_time is 10 and max_time is 50, the resulting histogram will include the following five buckets:

First bucket: 0 - 10
Second bucket: 10 - 20
Third bucket: 20 - 30
Fourth bucket: 30 - 50
Last bucket: 50 - infinity
This allows the user to see how the data is distributed within the full range of values, rather than just within the range of values specified by min_time and max_time.


EngineeredVirus:
Close as a different solution was implemented.

PR #350, opened 2023-01-09
PG-564: Added a column to display bind variables when used with the n…

kayform:
…ormalized_query option.

When collecting data for monitoring, display it as placeholders using the pgsm_normalized_query option.
If bind parameter and "normalized query are retrieved together as attached the image, it would be more useful for query performance analysis.


ibrarahmad (changes requested):
Please add some test cases for that.

ibrarahmad (changes requested):
pg_stat_monitor-1.0-2.0.sql need to be updated

AndersAstrand:
I'm not entirely sure I understand the purpose of this. Just as when `pgsm_normalized_query` is turned off, the values you will get is just for the first execution of the normalized query anyways. This column doesn't add any extra information above what `pgsm_normalized_query = off` will already give you, right?

Or am I completely misunderstanding what you want to accomplish here?

AndersAstrand:
Seems related to #481 and #517

AndersAstrand:
I'll close this as we seem to have lost the original creator when we just left it on read for 2 years.

PR #351, opened 2023-01-16
PG-579: Querying pg_stat_monitor crashes the server ...

codeforall:
pgsm_get_ss() must only be called when pg_stat_monitor.so is loaded. Fix is to move the pgsm_get_ss() call after checking if the pg_stat_monotor library is loaded or not.


EngineeredVirus (approved):
LGTM

PR #352, opened 2023-01-17
PG-543: pg_stat_monitor: PostgreSQL's pg_stat_statements compatible view.

EngineeredVirus:
The view now carries all the columns as pg_stat_statements. This required fixing data types of some of the columns, renaming a few, as well inclusion of new columns to make the view fully compatible with pg_stat_statements.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -1860,12 +1860,12 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 			nulls[i++] = true;
 
 		/* queryid at column number 4 */
-		values[i++] = Int64GetDatumFast(queryid);
+		values[i++] = UInt64GetDatum(queryid);
 
 		/* planid at column number 5 */
 		if (planid)
 		{
-			values[i++] = Int64GetDatumFast(planid);
+			values[i++] = UInt64GetDatum(planid);
 		}
 		else
 		{
@@ -1903,7 +1903,7 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 			values[i++] = CStringGetTextDatum("<insufficient privilege>");
 		}
 
-		values[i++] = Int64GetDatumFast(pgsm_query_id);
+		values[i++] = UInt64GetDatum(pgsm_query_id);
 
 		/* state at column number 8 for V1.0 API*/
 		if (api_version <= PGSM_V1_0)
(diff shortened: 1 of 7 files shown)


pg_stat_monitor.c
```diff
@@ -1863,14 +1860,12 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 			nulls[i++] = true;
 
 		/* queryid at column number 4 */
-		snprintf(queryid_text, 32, "%08lX", queryid);
-		values[i++] = CStringGetTextDatum(queryid_text);
+		values[i++] = Int64GetDatumFast(queryid);
```
codeforall, line 1863: queryid is declared as Int64, so use UInt64GetDatum  macro instead of Int64GetDatumFast
`values[i++] = UInt64GetDatum(queryid);`
Int64GetDatumFast will cause static assertion failure with the current head of the PG code base.

pg_stat_monitor.c
```diff
@@ -1863,14 +1860,12 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
...
 
 		/* planid at column number 5 */
 		if (planid)
 		{
-			snprintf(planid_text, 32, "%08lX", planid);
-			values[i++] = CStringGetTextDatum(planid_text);
+			values[i++] = Int64GetDatumFast(planid);
```
codeforall, line 1868: values[i++] = UInt64GetDatum(planid);

pg_stat_monitor.c
```diff
@@ -1917,8 +1912,7 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 		/* parentid at column number 9 */
 		if (tmp.info.parentid != UINT64CONST(0))
 		{
-			snprintf(parentid_txt, 32, "%08lX", tmp.info.parentid);
-			values[i++] = CStringGetTextDatum(parentid_txt);
+			values[i++] = Int64GetDatumFast(tmp.info.parentid);
```
codeforall, line 1915: change to 
`values[i++] = UInt64GetDatum(tmp.info.parentid);`

codeforall (changes requested):
PG source has added a static assertion to *GetDatum*() macros to validate the argument data type in the latest head. So I suggest we use Uint64 macros for Uint64 types

pg_stat_monitor.c
EngineeredVirus: Resolved with the latest commit.

pg_stat_monitor.c
EngineeredVirus: Resolved with the latest commit.

pg_stat_monitor.c
EngineeredVirus: Resolved with the latest commit.

ibrarahmad (approved):
LGTM

PR #353, opened 2023-01-17
PG-582: blk_read_time and blk_write_time are not being rounded.

EngineeredVirus:
Added the round off within the internal function so that values for blk_read_time, blk_write_time are rounded off to 4 decimal places.

Additionally, added rounding off for the PG15+ columns of temp_blk_read_time and temp_blk_write_time.


codeforall:
Although not mentioned in the Jira ticker, apart from the values rounded off by the pull request, four more float values are not getting rounded.

EngineeredVirus:
> 

Updated per your suggestion. Now the four JIT related columns are also being rounded off.

ibrarahmad (approved):
LGTM

PR #354, opened 2023-01-18
PG-581: top_queryid expected output verification and change.

Naeem-Akhter:
top_queryid test case was failing on all PG versions, pushing an update to expected output files.


codeforall (approved):
LGTM

PR #355, opened 2023-01-18
PG-584: Verify and 007_settings_pgsm_query_shared_buffer.pl TAP test

Naeem-Akhter:
Verified and updated the 007_settings_pgsm_query_shared_buffer test case where it was failing due to bucket change on PG 12~15 intermittently, and also reduce the execution time.


Naeem-Akhter:
I will squash another commit into this PR to add another file for PG-15. Please wait for that.

Naeem-Akhter:
Commits have been squashed. Please go ahead with review. @ibrarahmad @EngineeredVirus

EngineeredVirus (approved):
With the summation now being done in the queries, I don't think you need to increase the bucket time.

LGTM

PR #356, opened 2023-01-18
PG-400: pg_stat_monitor: Timezone in msgtime column...

codeforall:
The bucket start time reported by pg_stat_monitor does not match the PG time and timezone. The fix is to use TimestampTz for recording the bucket start time.


ibrarahmad (approved):
LGTM

PR #357, opened 2023-01-19
PG-562: Histogram Ranges/Buckets are not correct.

EngineeredVirus:
Added buckets for queries that take less than minimum histogram time and one for the ones taking more than the max value specified.

Also, in case the buckets end up overlapping, on server start, an error will be thrown informing the user of this issue and requesting a rectification.

Refactored the code to consolidate the calculations in a single function.


codeforall (approved):
Looks good

PR #358, opened 2023-01-20
PG-588: Some queries are not being normalised.

EngineeredVirus:
There is no specific test case where I can either reproduce or validate the fix. Though, one of the suspects is this condition in pgss_store. Therefore removed, and it requires verification.


codeforall (approved):
Looks good

PR #359, opened 2023-01-20
PG-586: pg_stat_monitor: CPU and user timing should be captured for utility statements as well

EngineeredVirus:
Added the necessary capture of resource usage  in the process utility function. We are now storing CPU and user timings for a utility statement.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -659,6 +659,7 @@ pgss_ExecutorEnd(QueryDesc *queryDesc)
 		 * levels of hook all do this.)
 		 */
 		InstrEndLoop(queryDesc->totaltime);
+
 		if (getrusage(RUSAGE_SELF, &rusage_end) != 0)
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -973,6 +980,13 @@ pgss_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 		}
 
 		PG_END_TRY();
+
+		if (getrusage(RUSAGE_SELF, &rusage_end) != 0)
+			elog(DEBUG1, "pg_stat_monitor: failed to execute getrusage");
+
+		sys_info.utime = time_diff(rusage_end.ru_utime, rusage_start.ru_utime);
```
codeforall, line 987: shouldn't we set sys_info.utime and sys_info.stime to 0 in case getrusage() return with error?

EngineeredVirus: Resolved. Made the changes suggested.

codeforall (approved):
Looks good

PR #362, opened 2023-01-23
PG-572: Verify 025_compare_pgss.pl TAP test case.

Naeem-Akhter:
1) Added the Group by clause to make sure that bucket change doesn't have any impact on aggregates of queries.
2) Updated column names where required.
3) Updated pgbench parameters to reduce the time taken by test case, around 70-80% decrease in time taken by test case.


EngineeredVirus (approved):
LGTM

PR #364, opened 2023-01-24
PG-597: pg_stat_monitor: Remove rounding off for floating point values

codeforall:
As an observability tool that serves data to other tools, data must be output without any loss. So rounding off causes data loss and rounding off errors when comparing different columns.

Therefore, it was decided to eliminate rounding off when outputting values. Any consumer of this data should round off data to whatever precision it prefers.

This behaviour is also consistent with pg_stat_statements.


EngineeredVirus (approved):
LGTM

PR #365, opened 2023-01-24
PG-572: Verify 025_compare_pgss.pl TAP test case.

Naeem-Akhter:
Updated test case for column name change (rows_retrieved -> rows).


codeforall (approved):
Looks good

PR #367, opened 2023-01-25
PG-601: Histogram ranges are not correct

EngineeredVirus:
Resolved the issue with histogram outlier buckets. Also updated the printing of bucket ranges to be in correct set notation with reference to brackets. The lower bounds of buckets always have an exclusive range except for the first bucket, and the upper bounds always have an inclusive value.
( or ) => exclusive
{ or } => inclusive

The entire range is enclosed within the {} brackets.


ibrarahmad (approved):
LGTM

PR #369, opened 2023-01-30
PG-559: Add a TAP testcase for historgram feature.

Naeem-Akhter:
Adding a tap test case for histogram feature where we are testing behaviour for different scenarios. 
1- Verify against default histogram values. 
2- Verify lower and upper range values, against user defined min and max. 
3- User defined histogram min, max and buckets. 
4- Negative values for min and max.


ibrarahmad (approved):
LGTM

PR #370, opened 2023-01-30
PG-587: pg_stat_monitor: Validate the upgrade from 1.x to 2.0 version

codeforall:
Disallow V1.0 API to be used with V2.0 lib and remove pg_stat_monitor--1.0.sql as part of that. A few adjustments to the 1.0 to 2.0 upgrade script are also part of the commit.


EngineeredVirus (approved):
LGTM

PR #379, opened 2023-02-22
PG-542: Performance improvement of pg_stat_monitor.

codeforall:
Saving the client IP address once per the lifetime of a backend. This avoid the expensive operation multiple times, and hence improving performance significantly.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -24,6 +24,8 @@
 #include "commands/dbcommands.h"
 #include "commands/explain.h"
 #include "pg_stat_monitor.h"
+#include "libpq/libpq-be.h"
+
 
  /*
   * Extension version number, for supporting older extension versions' objects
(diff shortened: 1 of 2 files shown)


pg_stat_monitor.c
```diff
@@ -1642,13 +1647,16 @@ pgsm_create_hash_entry(uint64 bucket_id, uint64 queryid, PlanInfo *plan_info)
...
 	pgsmEntry *entry;
 	int sec_ctx;
-	bool found_client_addr = false;
 	char app_name[APPLICATIONNAME_LEN] = "";
 	char *app_name_ptr = app_name;
 	int	app_name_len = 0;
 	MemoryContext oldctx;
 	char *datname = NULL;
 	char *username = NULL;
+	static uint	client_ip = 0xFFFFFFFF;
```
ibrarahmad, line 1656: Use global variable instead of static.

pg_stat_monitor.c
```diff
@@ -1281,32 +1283,35 @@ pg_get_application_name(char *name, int buff_size)
 	return strlen(name);
 }
 
+#include <string.h>
+#include <sys/errno.h>
+
 static uint
-pg_get_client_addr(bool *ok)
+get_client_ip_address_integer(void)
```
ibrarahmad, line 1290: Formatting Issue, too many whitespacess.

ibrarahmad (approved):
LGTM

PR #383, opened 2023-02-23
PG-606: New GUC required for enabling/disabling of pgsm_query_id calculation…

codeforall:
Adds a new GUC pg_stat_monitor.pgsm_enable_pgsm_query_id to enable/disable pgsm query id calculation. Apart from that patch also refactors the GUC-related code to match PostgreSQL conventions.

Moreover, the commit also changes the pgsm_enable_overflow GUC to boolean instead of enum.

### DIFF
--- a/guc.c
+++ b/guc.c
@@ -18,289 +18,262 @@
 
 #include "pg_stat_monitor.h"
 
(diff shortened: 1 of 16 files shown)


EngineeredVirus (approved):
LGTM

PR #384, opened 2023-02-23
PG-607: Allow histogram to track queries in sub-ms time brackets

codeforall:
Updated the GUC configuration and the relevant histogram functionality to track queries in lower cardinality than ms. This is done by saving the GUC values for histogram min and max values in real (double) type.

All test cases except for the 030 tap test are passing. The test case needs an update.


EngineeredVirus (approved):
LGTM.

The two test case failures are assigned to @Naeem-Akhter for review and fixes.

PR #386, opened 2023-02-23
PG-608: pg_stat_monitor: Update histogram TAP testcase for sub-ms.

Naeem-Akhter:
Due to enhancement in the histogram functionality to track queries in sub-ms time brackets, histogram TAP test case required changes to adapt to new behaviour. 

This PR also handles the failing SQL test case, GUC, on PG-11.


ibrarahmad (approved):
LGTM

PR #391, opened 2023-03-01
Multiple fixes including the addition of pgsm_query_id for errored out queries

EngineeredVirus:
This is an issue raised by the PMM team. Hence, we've also updated the error functions to store pgsm_query_id for errors.


codeforall (approved):
Looks good

PR #392, opened 2023-03-17
Make pg_config configurable in Makefile invocation

NyanKiyoshi:
This allows to override the `PG_CONFIG` variable through environment variables when wanting to use another one that is not in the `PATH`.


EngineeredVirus (approved):
Thank you for your contribution.

LGTM

PR #403, opened 2023-06-29
PG-592: Treat queries with different parent queries as separate entries

dutow:
PG-592

1. Previously pg_stat_monitor had a `topquery` and `topqueryid` field, but it was only a sample: it showed one of the top queries executing the specific query.

With this change, the same entry executed by two different functions will result in two entries in the statistics table.

2. This also fixes a bug where the content of these field disappeared for every second query executed: previously the update function changed topqueryid to `0` if it was non zero, and changed it to the proper id when it was 0. This resulted in an alternating behavior, where for every second executed query the top query disappeared.

After these changes, the top query is always shown.

3. The previous implementation also leaked dsa memory used to store the parent queries. This is now also fixed.


pg_stat_monitor.c
```diff
@@ -1900,6 +1911,8 @@ pgsm_store(pgsmEntry * entry)
 		shared_hash_entry->pgsm_query_id = entry->pgsm_query_id;
 		shared_hash_entry->encoding = entry->encoding;
 		shared_hash_entry->counters.info.cmd_type = entry->counters.info.cmd_type;
+		shared_hash_entry->counters.info.cmd_type = entry->counters.info.cmd_type;
```
codeforall, line 1914: The statement is redundant

pg_stat_monitor.c
```diff
@@ -20,6 +20,7 @@
 #include "nodes/pg_list.h"
 #include "utils/guc.h"
 #include <regex.h>
+#include <assert.h>
```
EngineeredVirus, line 23: This is not needed. See my comments below.

pg_stat_monitor.c
```diff
@@ -1472,11 +1473,7 @@ pgsm_update_entry(pgsmEntry * entry,
 			}
 			else
 			{
-				if(e->counters.info.parent_query != InvalidDsaPointer) {
-					dsa_area   *query_dsa_area = get_dsa_area_for_query_text();
-					dsa_free(query_dsa_area, e->counters.info.parent_query);
-					e->counters.info.parent_query = InvalidDsaPointer;
-				}
+				assert(e->counters.info.parent_query == InvalidDsaPointer);
```
EngineeredVirus, line 1476: Better to change this statement with:

Assert(DsaPointerIsValid(e->counters.info.parent_query) == false);

pg_stat_monitor.c
```diff
@@ -1438,9 +1439,9 @@ pgsm_update_entry(pgsmEntry * entry,
 			e->counters.info.num_relations = num_relations;
 			_snprintf2(e->counters.info.relations, relations, num_relations, REL_LEN);
 
-			if (exec_nested_level > 0 && e->key.parentid != 0 && pgsm_track == PGSM_TRACK_ALL)
+			if (exec_nested_level > 0 && exec_nested_level < max_stack_depth && e->key.parentid != 0 && pgsm_track == PGSM_TRACK_ALL)
 			{
-				if (e->counters.info.parent_query == InvalidDsaPointer && exec_nested_level >= 0 && exec_nested_level < max_stack_depth)
+				if (e->counters.info.parent_query == InvalidDsaPointer)
```
EngineeredVirus, line 1444: Just a thought:
IIUC, this statement is likely to return false most of the time. So, perhaps, this should be the first condition.

dutow: If the conditions above are false, this statement will always return true. (if we don't have a proper parent query, the pointer will be invalid)

pg_stat_monitor.c
dutow: Changed.

pg_stat_monitor.c
dutow: Fixed.

pg_stat_monitor.c
dutow: Fixed.

EngineeredVirus (approved):
LGTM

codeforall (approved):
Looks good

darkfronza (approved):
LGTM

Naeem-Akhter (approved):
LGTM!

PR #425, opened 2023-11-01
PG-645: pg_stat_monitor crashes PostgreSQL if there is citus ..

codeforall:
Do not look for the query in the hash if no query string is provided in the planner hook.


EngineeredVirus (approved):
LGTM

PR #429, opened 2023-11-11
Test PGSM with multiple extensions loaded in the server.

Naeem-Akhter:
Added a tap test case to load multiple extensions in the server before running a test load.

1. We load and create other extensions that are distributed by Percona in PPG. (postgis, pg_repack, pgaudit, pgaudit_set_user, pgpool)
2. Run test data with pgbench.

To make the above test case work, updated the workflows to install above mentioned extension where we use installed packages from PPG. On workflows where we build server or use packages from PGDG, we are skipping this test case.

Also removed the workflows for postgres 11, as it is already EOL.

Updated README.md to add pg-16 and remove pg-11.


t/031_multiple_extensions.pl
EngineeredVirus, line 1: I think we need to make this test case conditional, perhaps by using a define. It would otherwise make it very hard to run this test suite in a dev environment.

Naeem-Akhter: We already have this check in perl test case file where it is only run for installation with PPG package distribution.

codeforall (approved):
Looks good

PR #432, opened 2023-11-23
PG-646: pg_stat_monitor hangs in pgsm_store

EngineeredVirus:
A potential lock contention could've been caused when an OOM warning was being emitted by the pgsm_store function. This could lead the pg_store_error function calling pgsm_store function and thereby trying to acquire and exclusive lock when a shared was already by the same process. This warning is now guarded by protection block.


codeforall (approved):
Looks good

PR #433, opened 2023-11-28
Add files via upload

ksjj97inzent:
add bind_variables ver 2.03


ksjj97inzent:
ormalized_query option.

When collecting data for monitoring, display it as placeholders using the pgsm_normalized_query option.
If bind parameter and "normalized query are retrieved together as attached the image, it would be more useful for query performance analysis.

add bind_variables ver 2.03

Issue #412, opened 2023-07-17
DB does not start after setting pgsm_max to its maximum


dutow:
The pg_stat_monitor.pgsm_max should accept up to 10240MB. When setting it to the maximum, the DB does not start anymore. It seems to work with the maximum - 1MB

https://jira.percona.com/browse/PG-632

```
[root@pl242-pablo-svampa-default ~]# psql --pset pager -c "SELECT name, setting, unit, min_val, max_val, short_desc FROM pg_settings WHERE name like 'pg_stat_monitor.pgsm_max'"
           name           | setting | unit | min_val | max_val |                                                short_desc
--------------------------+---------+------+---------+---------+----------------------------------------------------------------------------------------------------------
 pg_stat_monitor.pgsm_max | 10      | MB   | 10      | 10240   | Sets the maximum size of shared memory in (MB) used for statement's metadata tracked by pg_stat_monitor.
(1 row)
[root@pl242-pablo-svampa-default ~]# psql --pset pager -c "ALTER SYSTEM SET pg_stat_monitor.pgsm_max = 10239"
ALTER SYSTEM
[root@pl242-pablo-svampa-default ~]# systemctl restart postgresql-14.service
[root@pl242-pablo-svampa-default ~]# psql --pset pager -c "SELECT name, setting, unit, min_val, max_val, short_desc FROM pg_settings WHERE name like 'pg_stat_monitor.pgsm_max'"
           name           | setting | unit | min_val | max_val |                                                short_desc
--------------------------+---------+------+---------+---------+----------------------------------------------------------------------------------------------------------
 pg_stat_monitor.pgsm_max | 10239   | MB   | 10      | 10240   | Sets the maximum size of shared memory in (MB) used for statement's metadata tracked by pg_stat_monitor.
(1 row)
[root@pl242-pablo-svampa-default ~]# psql --pset pager -c "ALTER SYSTEM SET pg_stat_monitor.pgsm_max = 10240"
ALTER SYSTEM
[root@pl242-pablo-svampa-default ~]# systemctl restart postgresql-14.service
Job for postgresql-14.service failed because the control process exited with error code. See "systemctl status postgresql-14.service" and "journalctl -xe" for details.
[root@pl242-pablo-svampa-default ~]#
```

```
[...]

PR #435, opened 2024-01-30
Fix integer overflow

artemgavrilov:
Fixes https://github.com/percona/pg_stat_monitor/issues/412

This PR fixes 3 integer overflow bugs, all of them covered with tests.


pg_stat_monitor.c
```diff
@@ -1999,7 +1999,7 @@ IsBucketValid(uint64 bucketid)
 
 	TimestampDifference(pgsm->bucket_start_time[bucketid], current_tz, &secs, &microsecs);
 
-	if (secs > (pgsm_bucket_time * pgsm_max_buckets))
+	if (secs > ((long long)pgsm_bucket_time * pgsm_max_buckets))
```
artemgavrilov, line 2002: `pgsm_bucket_time` max value is `INT_MAX` and `pgsm_max_buckets` can be up to `20000`

t/031_query_stat.pl
```diff
@@ -0,0 +1,63 @@
...
+
+# CREATE new PostgreSQL node and do initdb
+my $node = PGSM->pgsm_init_pg();
+my $pgdata = $node->data_dir;
+
+# UPDATE postgresql.conf to include/load pg_stat_monitor library
+open my $conf, '>>', "$pgdata/postgresql.conf";
+print $conf "shared_preload_libraries = 'pg_stat_monitor'\n";
+print $conf "pg_stat_monitor.pgsm_bucket_time = 2147483647\n"; # Max value for this parameter 
+print $conf "pg_stat_monitor.pgsm_max_buckets = 20000\n"; # Max value for this parameter
```
artemgavrilov, line 23: This combination of parameters caused int overflow in `IsBucketValid` function.

pg_stat_monitor.c
Naeem-Akhter: TAP test cases failing with following: 

# All 17 tests passed.
echo "# +++ tap install-check in  +++" && rm -rf '/home/postgres/src/pg_stat_monitor'/tmp_check && /bin/mkdir -p '/home/postgres/src/pg_stat_monitor'/tmp_check && cd ./ && TESTLOGDIR='/home/postgres/src/pg_stat_monitor/tmp_check/log' TESTDATADIR='/home/postgres/src/pg_stat_monitor/tmp_check' PATH="/usr/lib/postgresql/16/bin:/home/postgres/src/pg_stat_monitor:$PATH" PGPORT='65432' PG_REGRESS='/usr/lib/postgresql/16/lib/pgxs/src/makefiles/../../src/test/regress/pg_regress' /usr/bin/prove -I /usr/lib/postgresql/16/lib/pgxs/src/makefiles/../../src/test/perl/ -I ./  t/*.pl
# +++ tap install-check in  +++
t/001_settings_default.pl ................... ok
t/002_settings_pgsm_track_planning.pl ....... ok
t/003_settings_pgms_extract_comments.pl ..... ok
t/004_settings_pgsm_track.pl ................ ok
t/005_settings_pgsm_enable_query_plan.pl .... ok
t/006_settings_pgsm_overflow_target.pl ...... ok
t/007_settings_pgsm_query_shared_buffer.pl .. 17/? Bailout called.  Further testing stopped:  command "pg_ctl -w -D /home/postgres/src/pg_stat_monitor/tmp_check/t_007_settings_pgsm_query_shared_buffer_pgsm_regression_data/pgdata -l /home/postgres/src/pg_stat_monitor/tmp_check/log/007_settings_pgsm_query_shared_buffer_pgsm_regression.log restart" exited with value 1
FAILED--Further testing stopped: command "pg_ctl -w -D /home/postgres/src/pg_stat_monitor/tmp_check/t_007_settings_pgsm_query_shared_buffer_pgsm_regression_data/pgdata -l /home/postgres/src/pg_stat_monitor/tmp_check/log/007_settings_pgsm_query_shared_buffer_pgsm_regression.log restart" exited with value 1
make: *** [/usr/lib/postgresql/16/lib/pgxs/src/makefiles/pgxs.mk:437: installcheck] Error 255

artemgavrilov: @Naeem-Akhter I wasn't able to reproduce this issue on my local setup (MacOS). I need more time to investigate it.

artemgavrilov: @Naeem-Akhter problem was in the small amount of memory. I tuned tests to fit in 2GB of memory, but still check overflows. However I removed tests that check maximum values for `pgsm_max` and `pgsm_query_shared_buffer` because it will require more than 10GB of memory.

Here is Vagrantfile that I used for tests:
```
Vagrant.configure("2") do |config|
  config.vm.box = "bento/ubuntu-22.04-arm64"
  config.vm.provider :vmware_desktop do |vmware|
    vmware.gui = false
    vmware.cpus = 2
    vmware.memory = "2048"
    vmware.vmx["ethernet0.virtualdev"] = "vmxnet3"
    vmware.ssh_info_public = true
    vmware.linked_clone = false
  end
end
```

Naeem-Akhter: Makes sense !! :)

Naeem-Akhter (approved):
Looks good to me!

Naeem-Akhter:
@codeforall @hqakhtar @dutow @dAdAbird 

Can one of your please review c code files changes in this PR?

pg_stat_monitor.c
dAdAbird, line 2002: I'd rather use a PG's typedef `int64` instead of `long long` to be consistent with the rest of the code

pg_stat_monitor.h
```diff
@@ -93,8 +93,8 @@
 /* the assumption of query max nested level */
 #define DEFAULT_MAX_NESTED_LEVEL	10
 
-#define MAX_QUERY_BUF						(pgsm_query_shared_buffer * 1024 * 1024)
-#define MAX_BUCKETS_MEM 					(pgsm_max * 1024 * 1024)
+#define MAX_QUERY_BUF						((long long)pgsm_query_shared_buffer * 1024 * 1024)
+#define MAX_BUCKETS_MEM 					((long long)pgsm_max * 1024 * 1024)
```
dAdAbird, line 97: Logically it has to be `size_t` (or PG's `Size`?) but `pgsm_query_shared_buffer` and `pgsm_max` are signed? Should they be signed? @Naeem-Akhter, @codeforall

pg_stat_monitor.c
artemgavrilov: Done

pg_stat_monitor.h
artemgavrilov: These variables are GUC variables, and according to documentation only `boolean`, `enum`, `int`, `real`, and `string` types are supported. 

https://github.com/postgres/postgres/blob/master/src/backend/utils/misc/README#L3-L9

dAdAbird (approved):

PR #438, opened 2024-03-28
Fix IPC::Run perl module name in CI

artemgavrilov:
```
Warning: Cannot install IPC::RUN, don't know what it is.
Try the command

    i /IPC::RUN/

to find objects with matching identifiers.
```

https://metacpan.org/pod/IPC::Run

Same for pg_tde: https://github.com/Percona-Lab/pg_tde/pull/156

### DIFF
--- a/.github/workflows/postgresql-12-pgdg-package.yml
+++ b/.github/workflows/postgresql-12-pgdg-package.yml
@@ -23,7 +23,7 @@ jobs:
             /usr/lib/postgresql /usr/include/postgresql /usr/share/postgresql \
             /etc/postgresql
           sudo rm -f /usr/bin/pg_config
-          sudo /usr/bin/perl -MCPAN -e 'install IPC::RUN'
+          sudo /usr/bin/perl -MCPAN -e 'install IPC::Run'
           sudo /usr/bin/perl -MCPAN -e 'install Text::Trim'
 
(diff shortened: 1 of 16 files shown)


.github/workflows/postgresql-12-pgdg-package.yml
```diff
@@ -23,7 +23,7 @@ jobs:
             /usr/lib/postgresql /usr/include/postgresql /usr/share/postgresql \
             /etc/postgresql
           sudo rm -f /usr/bin/pg_config
-          sudo /usr/bin/perl -MCPAN -e 'install IPC::RUN'
+          sudo /usr/bin/perl -MCPAN -e 'install IPC::Run'
```
artemgavrilov, line 26: We installing perl modules in `Delete old postgresql files` step, that looks wrong. Should we extract these actions to separate step?

artemgavrilov:
CI jobs that are failed it's same set that fails on main branch.

Naeem-Akhter (approved):
Looks Good!

.github/workflows/postgresql-12-pgdg-package.yml
Naeem-Akhter: No harm in doing it. But as these are sequential steps so either way we are likely to see same behaviour.

dAdAbird (approved):

PR #441, opened 2024-04-02
Fix cppcheck workflow

artemgavrilov:
`cppcheck` version available in ubuntu repos is pretty old. `cppcheck` doesn't provide pre-build binaries, so it looks like the best option to get fresh version is to build it from sources. Version that we used had some bugs that were fixed in recent releases, so just updating it healed workflow.


.github/workflows/cppcheck.yml
```diff
@@ -4,12 +4,27 @@ on: [push]
 jobs:
   build:
     name: cppcheck-test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
artemgavrilov, line 7: Just to use same ubuntu in all workflows

Naeem-Akhter (approved):
LGTM.

dAdAbird (approved):

PR #444, opened 2024-04-08
Tune CI triggers

artemgavrilov:
This PR sets same CI triggers for all workflows. It triggers all workflows on push in `main` branch, push of release tags and in Pull Requests. With that change we will be able to compare state of CI in feature PRs with something stable (main branch or release). Now it's hard sometimes to understand if it's your changes caused problems in CI or same problem exists in `main`


Naeem-Akhter:
I would suggest to have this discussed in next full-force DS, as earlier we decided to have some workflows specific to PRs and others to PUSH. 

Maybe we need to exclude the workflows against the Server Build too, as these are time consuming and potentially an unneeded overhead. And just adopt the server install via packages workflows only for both PR and PUSH.

Naeem-Akhter (approved):
Looks good to me.

PR #445, opened 2024-04-09
Fix workflows after CI image upgrade

artemgavrilov:
After upgrade CI image to `ubuntu-22.04` some workflows start failing. Fix was copied from `tde` repo: https://github.com/Percona-Lab/pg_tde/blob/38ff41d7b21efb62502978a31e8e378ca3f279f6/.github/workflows/postgresql-16-pgdg-package-pgxs.yml#L50-L53


codeforall (approved):
LGTM

PR #456, opened 2024-04-18
Added a tap test case to load multiple PPG extensions in the server before running a test load.

Naeem-Akhter:
Added a tap test case to load multiple PPG extensions in the server before running a test load.

1. We load and create other extensions that are distributed by Percona in PPG. (postgis, pg_repack, pgaudit, pgaudit_set_user, pgpool)
2. Run test data with pgbench.

To make the above test case work, updated the workflows to install above mentioned extension where we use installed packages from PPG. On workflows where we build server or use packages from PGDG, we are skipping this test case.


artemgavrilov:
@Naeem-Akhter  It looks like in this PR you accidentally changed permissions for `t/expected/007_settings_pgsm_query_shared_buffer.out.15` file. 

Screenshot 2024-04-19 at 12 40 19

Naeem-Akhter:
> @Naeem-Akhter It looks like in this PR you accidentally changed permissions for `t/expected/007_settings_pgsm_query_shared_buffer.out.15` file.
> 
> Screenshot 2024-04-19 at 12 40 19

No, it is intentional, to make it consistent with rest of the files.

artemgavrilov (approved):

PR #469, opened 2024-07-02
[PG-644] Add option to disable application name tracking

artemgavrilov:
PG-644

This PR introduces two changes:

1. It uses new functions for PG 16 and 17 for getting  backend status that allow to avoid iteration over be status table.
2. It adds new parameter `pgsm_track_application_names` that allows to disable application name tracking

It turned out that taking `application name` is expensive operation, it requires locking inside PG. This causes performance dramatically on big amount of connections. For example on TPC-C benchmark with 512 threads PGSM with application name tracking (default) gives -15% QPS, while without tracking only -3% QPS.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+# https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners
+# Order is important; the last matching pattern takes the most precedence.
+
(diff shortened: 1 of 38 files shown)


Naeem-Akhter (approved):
LGTM.

codeforall (approved):
Looks good.

darkfronza (approved):
LGTM

PR #475, opened 2024-07-17
Format sources

artemgavrilov:
This PR adds Makefile targets that format sources according to PG rules and CI workflow that checks that opened pull request has formatted sources (CI fails if sources are not formatted). 

https://wiki.postgresql.org/wiki/Running_pgindent_on_non-core_code_or_development_code
https://github.com/postgres/postgres/blob/master/src/tools/pgindent/README


darkfronza (approved):
LGTM

codeforall (approved):

PR #481, opened 2024-08-01
PG -156: replace query placeholders with actual arguments for prepared statements

darkfronza:
[PG-156](https://perconadev.atlassian.net/browse/PG-156)

Before this PR, `pg_stat_monitor` was unable to replace query placeholders when tracking prepared statements if `pg_stat_monitor.pgsm_normalized_query` is `off`.

With this update `pg_stat_monitor` now correctly store query argument values for prepared statements.

[PG-156]: https://perconadev.atlassian.net/browse/PG-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -208,13 +208,8 @@ static void pgsm_add_to_list(pgsmEntry *entry, char *query_text, int query_len);
 static pgsmEntry *pgsm_get_entry_for_query(uint64 queryid, PlanInfo *plan_info, const char *query_text, int query_len, bool create);
 static uint64 get_pgsm_query_id_hash(const char *norm_query, int len);
 
-/*
- * extract parameter value (Datum) from plist->params[idx], cast it to string then
- * append the resulting string to the buffer.
- */
 static void get_param_value(const ParamListInfo plist, int idx, StringInfoData *buffer);
 
-/* denormalize the query, replace placeholders with actual values */
 static StringInfoData get_denormalized_query(const ParamListInfo paramlist, const char *query_text);
 
 static void pgsm_cleanup_callback(void *arg);
@@ -243,7 +238,13 @@ static void pgsm_update_entry(pgsmEntry *entry,
 							  const struct JitInstrumentation *jitusage,
 							  bool reset,
 							  pgsmStoreKind kind);
-static void pgsm_store(pgsmEntry *entry);
+static void pgsm_store_ex(pgsmEntry * entry, ParamListInfo params);
+
+/* Stores query entry in normalized form */
+static inline void pgsm_store(pgsmEntry * entry)
+{
+	pgsm_store_ex(entry, NULL);
+}
 
 static void pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 									 pgsmVersion api_version,
@@ -703,7 +704,6 @@ pgsm_ExecutorEnd(QueryDesc *queryDesc)
 	PlanInfo	plan_info;
 	PlanInfo   *plan_ptr = NULL;
 	pgsmEntry  *entry = NULL;
-	StringInfoData query_info;
 	MemoryContext oldctx;
 
 	/* Extract the plan information in case of SELECT statement */
@@ -763,13 +763,6 @@ pgsm_ExecutorEnd(QueryDesc *queryDesc)
 			sys_info.stime = time_diff(rusage_end.ru_stime, rusage_start.ru_stime);
 		}
 
-        if (!pgsm_normalized_query && queryDesc->params) {
-            query_info = get_denormalized_query(queryDesc->params, queryDesc->sourceText);
-            oldctx = MemoryContextSwitchTo(GetPgsmMemoryContext());
-            entry->query_text.query_pointer = pnstrdup(query_info.data, query_info.len);
-            MemoryContextSwitchTo(oldctx);
-        }
-
 		pgsm_update_entry(entry,	/* entry */
 						  NULL, /* query */
 						  NULL, /* comments */
@@ -794,7 +787,7 @@ pgsm_ExecutorEnd(QueryDesc *queryDesc)
 						  false,	/* reset */
 						  PGSM_EXEC);	/* kind */
 
-		pgsm_store(entry);
+		pgsm_store_ex(entry, queryDesc->params);
 	}
 
 	if (prev_ExecutorEnd)
@@ -1880,7 +1873,6 @@ pgsm_create_hash_entry(uint64 bucket_id, uint64 queryid, PlanInfo *plan_info)
 	return entry;
 }
 
-
 /*
  * Store some statistics for a statement.
  *
@@ -1890,9 +1882,14 @@ pgsm_create_hash_entry(uint64 bucket_id, uint64 queryid, PlanInfo *plan_info)
  * If jstate is not NULL then we're trying to create an entry for which
  * we have no statistics as yet; we just want to record the normalized
  * query string.  total_time, rows, bufusage are ignored in this case.
+ *
+ * If params argument is not null and pgsm_normalized_query is off then we
+ * denormalize the query using it's actual arguments found in params.
+ * The denormalization is done during the first time the query is
+ * inserted or if the time to execute the query exceeds the average
+ * time computed for the same query.
  */
-static void
-pgsm_store(pgsmEntry *entry)
+void pgsm_store_ex(pgsmEntry *entry, ParamListInfo params)
 {
 	pgsmEntry  *shared_hash_entry;
 	pgsmSharedState *pgsm;
@@ -1907,6 +1904,7 @@ pgsm_store(pgsmEntry *entry)
 	JitInstrumentation jitusage;
 	char		comments[COMMENTS_LEN] = {0};
 	int			comments_len;
+	StringInfoData query_info;
 
 	/* Safety check... */
 	if (!IsSystemInitialized())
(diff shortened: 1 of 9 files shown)


pg_stat_monitor.c
```diff
@@ -206,6 +208,15 @@ static void pgsm_add_to_list(pgsmEntry *entry, char *query_text, int query_len);
...
 
+/*
+ * extract parameter value (Datum) from plist->params[idx], cast it to string then
+ * append the resulting string to the buffer.
+ */
+static void get_param_value(const ParamListInfo plist, int idx, StringInfoData *buffer);
+
+/* denormalize the query, replace placeholders with actual values */
+static StringInfoData get_denormalized_query(const ParamListInfo paramlist, const char *query_text);
+
```
artemgavrilov, line 219: Other functions in this file have their documentation comments on implementations, not on prototypes. I don't have any preference, but better to keep consistency.

pg_stat_monitor.c
```diff
@@ -232,7 +243,17 @@ static void pgsm_update_entry(pgsmEntry *entry,
 							  const struct JitInstrumentation *jitusage,
 							  bool reset,
 							  pgsmStoreKind kind);
-static void pgsm_store(pgsmEntry *entry);
+/*
+ * Pass the query arguments so it can be denormalized if
+ * pgsm_normalized_query is off.
+ */
+static void pgsm_store_ex(pgsmEntry * entry, ParamListInfo params);
```
artemgavrilov, line 250: I wonder if we can just make `params` part of `entry`? Does it makes sense?

artemgavrilov: BTW `_ex` stands for `extended`?

pg_stat_monitor.c
```diff
@@ -2064,6 +2097,43 @@ pgsm_store(pgsmEntry *entry)
 		snprintf(shared_hash_entry->datname, sizeof(shared_hash_entry->datname), "%s", entry->datname);
 		snprintf(shared_hash_entry->username, sizeof(shared_hash_entry->username), "%s", entry->username);
 	}
+	/* Entry already exists, if query normalization is disabled and
+	 * the query execution time exceeds the mean time for this query,
+	 * then we denormalize the query so users can inspect which arguments
+	 * caused the query to take more time to execute */
```
artemgavrilov, line 2103: So we always store most `expensive` parameters, am I right?

pg_stat_monitor.c
```diff
@@ -4018,3 +4088,84 @@ get_query_id(JumbleState *jstate, Query *query)
...
+		// append query text before the '$' sign found.
+		if (len > 0)
+			appendBinaryStringInfo(&result_buf, cursor_curr, len);
+
+		// skip '$'
+		++cursor_ori;
+
+		/* skip the placeholder */
+		while(*cursor_ori && *cursor_ori >= '0' && *cursor_ori <= '9')
+			cursor_ori++;
```
artemgavrilov, line 4158: Is there any reason why you are using `++cursor_ori` in one place and `cursor_ori++` in another?

pg_stat_monitor.c
```diff
@@ -4018,3 +4088,84 @@ get_query_id(JumbleState *jstate, Query *query)
...
+
+		// advance current cursor
+		cursor_curr = cursor_ori;
+
+		/* replace the placeholder with actual value */
+		get_param_value(paramlist, current_param, &result_buf);
+
+		++current_param;
+	} while (*cursor_ori != '\0');
```
artemgavrilov, line 4167: On a sting `4140` we are checking string termination like `!*cursor_ori` and here like `*cursor_ori != '\0'`. While both of them are equivalent as far as I know, should we stick to one approach? And if no what is the general rule where to use each of them?

regression/expected/pgsm_query_id.out
```diff
@@ -87,8 +87,8 @@ SELECT datname, pgsm_query_id, query, calls FROM pg_stat_monitor ORDER BY pgsm_q
  db1                | 1897482803466821995 | SELECT * FROM t2                                    |     3
  db1                | 1988437669671417938 | SELECT * FROM t1                                    |     1
  db2                | 1988437669671417938 | SELECT * FROM t1                                    |     1
- db1                | 2864453209316739369 | select $1 + $2                                      |     1
- db2                | 2864453209316739369 | select $1 + $2                                      |     1
+ db1                | 2864453209316739369 | select 1 + 2                                        |     1
+ db2                | 2864453209316739369 | select 1 + 2                                        |     1
```
artemgavrilov, line 91: I'd prefer to use some other values in ADD fuction, like 3 and 4, just to make it clear that it has nothing to do with parameter placeholders. At first sight it looked like we just dropped `$` sign here.

regression/sql/denorm_prepared_statements.sql
```diff
@@ -0,0 +1,19 @@
...
+Set pg_stat_monitor.pgsm_normalized_query='off';
+
+CREATE TABLE t1 (a TEXT, b TEXT, c TEXT);
+
+SELECT pg_stat_monitor_reset();
+
+PREPARE prepstmt(TEXT, TEXT, TEXT) AS INSERT INTO t1(a, b, c) VALUES($1, $2, $3);
+
+EXECUTE prepstmt('A', 'B', 'C');
+SELECT query, calls FROM pg_stat_monitor ORDER BY query COLLATE "C";
```
artemgavrilov, line 11: Just to make it clear that we are checking exact same data:
```suggestion
SELECT SUBSTRING(query, 0, 128), calls FROM pg_stat_monitor ORDER BY query COLLATE "C";
```

regression/sql/denorm_prepared_statements.sql
```diff
@@ -0,0 +1,19 @@
...
+
+EXECUTE prepstmt(REPEAT('XYZ', 8192), md5(random()::text), REPEAT('RANDOM', 4096));
+SELECT SUBSTRING(query, 0, 128), calls FROM pg_stat_monitor ORDER BY query COLLATE "C";
+
+DROP TABLE t1;
+
```
artemgavrilov, line 17: We didn't check case when at first we have heavy query and then cheap executed.

pg_stat_monitor.c
```diff
@@ -4018,3 +4088,84 @@ get_query_id(JumbleState *jstate, Query *query)
...
+	char *pstring;
+	ParamExternData *param;
+
+	Assert(idx < plist->numParams);
+
+	param = &plist->params[idx];
+
+	if (param->isnull || !OidIsValid(param->ptype))
+	{
+		appendStringInfoString(buffer, "NULL");
```
artemgavrilov, line 4106: Maybe it's better to keep placeholders? Wouldn't it be confusing if real NULL was used or we just replaced unknown parameter with NULL?

pg_stat_monitor.c
darkfronza: Yes, we can improve this later, like implementation details documented in the implementation, and API usage documented in the declaration. fixup: https://github.com/percona/pg_stat_monitor/pull/481/commits/f871606fb7da38b31340771aec7505005d7d867d

pg_stat_monitor.c
darkfronza: I prefer to keep this more explicit, as is, otherwise will need to ensure consistency of this new attribute usage/initialization in every entry usage in code so we don't end trying to denormalize an invalid pointer.

pg_stat_monitor.c
darkfronza: Yes, the "most expensive" query definition here is the one whose total execution time took more than the average execution time computed so far.

pg_stat_monitor.c
darkfronza: No, on a line by itself prefix/postfix doesn't make any difference that's why I like Go, it made available only postfix ++ and it's a statement, not an expression, to avoid confusion.
Postfix could be more expensive as there could be a temporary variable allocated to store the value returned by the expression before incrementing the variable, it's usually worst in c++ whose ++ operator can be applied to a big object, but either way -O2 optmizes this if the returned value is not used.

pg_stat_monitor.c
darkfronza: Usually I would write just `if (*ptr)` instead of `if (*ptr != '\0')` as \0 is zero and in the end we just compare if the expression is zero or not. I think due to the original patch I ended up mixing code.
The comparison with '\0' makes it more explicit that we are dealing with a `char*` and are looking for the end of the string, it's a matter of preference.

regression/expected/pgsm_query_id.out
darkfronza: Good idea, update on https://github.com/percona/pg_stat_monitor/pull/481/commits/f8d547451237fb31a72449eb0bcdeb56389c7cfd

regression/sql/denorm_prepared_statements.sql
darkfronza: Good, fixup on https://github.com/percona/pg_stat_monitor/pull/481/commits/20dd640b1e935fb35daaf65148e9e6104617cb3c

regression/sql/denorm_prepared_statements.sql
darkfronza: Added a new test case for this: https://github.com/percona/pg_stat_monitor/pull/481/commits/11e0cd539cf05217e7d4d02269fb8b16c9a2ae5a

pg_stat_monitor.c
darkfronza: if real NULL was used them `param->isnull` is true.

artemgavrilov (approved):

codeforall (approved):

PR #507, opened 2025-01-24
Add OSSF best practices badge

artemgavrilov:
PG-1238


jeltz (approved):
Simple enough change. The link is correct and says that is indeed compliant.

PR #515, opened 2025-02-14
PG-1349 Fix deadlock

artemgavrilov:
PG-1349

### Description

This PR fixed deadlock problem that was caused by re-acquire attempt of LWLock in case of erreport raise in pgsm_store function. When pgsm_store function throws an exception after acquiring LWLock it doesn't released automatically. `ereport` function uses longjump underhood to jump back to main loop, where is invokes error hooks. We have our own error hook that logs error message back to the map protected by the same LWLock. So that is the place where we got deadlock. Solution proposed by this PR is to wrap all code after acquiring lock with try-catch construction and forcefully release lock in case of error. That required to `InterruptHoldoffCount` counter and PG uses it to track code segments that should not be interrupted by signals. This approach described in errfinish function in elog.c

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -1968,120 +1968,142 @@ pgsm_store(pgsmEntry *entry)
 	 * we need to create the entry.
 	 */
 	LWLockAcquire(pgsm->lock, LW_SHARED);
-	shared_hash_entry = (pgsmEntry *) pgsm_hash_find(get_pgsmHash(), &entry->key, &found);
 
-	if (!shared_hash_entry)
+	PG_TRY();
 	{
-		dsa_pointer dsa_query_pointer;
-		dsa_area   *query_dsa_area;
-		char	   *query_buff;
 
-		/* New query, truncate length if necessary. */
-		if (query_len > pgsm_query_max_len)
-			query_len = pgsm_query_max_len;
+		shared_hash_entry = (pgsmEntry *) pgsm_hash_find(get_pgsmHash(), &entry->key, &found);
 
-		/* Save the query text in raw dsa area */
-		query_dsa_area = get_dsa_area_for_query_text();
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -1968,120 +1968,142 @@ pgsm_store(pgsmEntry *entry)
...
+					{
+						ereport(WARNING,
+								(errcode(ERRCODE_OUT_OF_MEMORY),
+								 errmsg("[pg_stat_monitor] pgsm_store: Hash table is out of memory and can no longer store queries!"),
+								 errdetail("You may reset the view or when the buckets are deallocated, pg_stat_monitor will resume saving " \
+										   "queries. Alternatively, try increasing the value of pg_stat_monitor.pgsm_max.")));
+					} PGSM_END_DISABLE_ERROR_CAPTURE();
+				}
+
+				LWLockRelease(pgsm->lock);
```
artemgavrilov, line 2047: This LWLockRelease was moved under `ereport` invocation to prevent double release in case of error.

pg_stat_monitor.c
```diff
@@ -1968,120 +1968,142 @@ pgsm_store(pgsmEntry *entry)
...
+		 * again in pgsm_emit_log_hook
+		 *
+		 * On ereport with level >= ERROR Postgres resets
+		 * InterruptHoldoffCount counter (see errfinish funcion in elog.c). We
+		 * need to increment it to be able to release lock here.
+		 */
+		HOLD_INTERRUPTS();		/* Increments holdoff counter */
+		LWLockRelease(pgsm->lock);	/* Decrements holdoff counter */
+		PG_RE_THROW();
+	}
```
artemgavrilov, line 2104: This is a fix itself. I wrapped all pgsm_store function body in try-catch to actually catch all possible ereports and force lock release here. For that we have to restore InterruptHoldoffCount counter because lock/release operations track it and assert against it.

artemgavrilov:
Closed in favor of https://github.com/percona/pg_stat_monitor/pull/518

PR #517, opened 2025-02-17
Revert "PG-156: replace query placeholders with actual arguments for…

artemgavrilov:
… prepared statements (#481)"

This reverts commit c921d483a8238bac7122566f9f705c0a09ab46dc.

PG-156

### Description

We discovered some bugs that blocks recent release. So we decided to revert this improvement and move the ticket back to development.


jeltz (changes requested):
There should be an explanation, ideally in both the pull request and the commit, for why this change was reverted. If we run into a problem in the future it would be very useful to be able to understand why.

artemgavrilov:
> There should be an explanation, ideally in both the pull request and the commit, for why this change was reverted. If we run into a problem in the future it would be very useful to be able to understand why.

Added short explanation to the PR description (it will became commit description after squash)

jeltz (approved):
Good enough description for me.

PR #518, opened 2025-02-20
PG-1349 Prevent LWLock deadlocks from happening

jeltz:
Instead of trying to fix every case where we could throw an error and handling that properly we just make sure to disable the error capture of the hook while our backend holds the lock.

We keep the check for `IsSystemOOM()` in the hook even though that might not be relevant anymore because if we are in OOM it is not like there would be any point to log the error anyway.

This is done via a global variable, similar to the `__pgsm_do_not_capture_error` variable that we are replacing, which we also use in one place to disable recursive calls to the log hook where we do not hold the lock.

A potential future improvement would be to make this variable a counter, or have two separate globals, so that we could guard against recursive calls to the hook running us out of stack and not just prevent the deadlocks.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -262,6 +262,18 @@ static int	comp_location(const void *a, const void *b);
 static uint64 get_next_wbucket(pgsmSharedState *pgsm);
 
 /*
+ * To prevent deadlocks against our own backend we need to disable error
+ * capture while holding the LWLock. The error capture hook is resposible
+ * itself for re-enabling data capture when called on ERROR or above since
+ * then we may not have been able to call pgsm_lock_release() due to the
+ * statement being aborted.
+ */
+static bool disable_error_capture = false;
+
+static void pgsm_lock_aquire(pgsmSharedState *pgsm, LWLockMode mode);
+static void pgsm_lock_release(pgsmSharedState *pgsm);
+
+/*
  * Module load callback
  */
 /*  cppcheck-suppress unusedFunction */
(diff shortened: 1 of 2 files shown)


pg_stat_monitor.c
```diff
@@ -1996,30 +2008,27 @@ pgsm_store(pgsmEntry *entry)
...
-				shared_hash_entry = hash_entry_alloc(pgsm, &entry->key, GetDatabaseEncoding());
-			}
-			PG_CATCH();
-			{
-				LWLockRelease(pgsm->lock);
+			shared_hash_entry = hash_entry_alloc(pgsm, &entry->key, GetDatabaseEncoding());
+		}
+		PG_CATCH();
+		{
+			pgsm_lock_release(pgsm);
```
artemgavrilov, line 2021: Releasing lock in catch section will violate assertion inside `LWLockRelease` function. When exception raised PG resets `InterruptHoldoffCount`, see https://github.com/percona/postgres/blob/e6cf26957f70367f84fbf5e16972634115decb22/src/backend/utils/error/elog.c#L519-L532.

So to be able to release lock here we have to increment the counter first with `HOLD_INTERRUPTS()`

pg_stat_monitor.c
```diff
@@ -2032,14 +2041,13 @@ pgsm_store(pgsmEntry *entry)
...
-							 errdetail("You may reset the view or when the buckets are deallocated, pg_stat_monitor will resume saving " \
-									   "queries. Alternatively, try increasing the value of pg_stat_monitor.pgsm_max.")));
-				} PGSM_END_DISABLE_ERROR_CAPTURE();
+				disable_error_capture = true;
+				ereport(WARNING,
+						(errcode(ERRCODE_OUT_OF_MEMORY),
+						 errmsg("[pg_stat_monitor] pgsm_store: Hash table is out of memory and can no longer store queries!"),
+						 errdetail("You may reset the view or when the buckets are deallocated, pg_stat_monitor will resume saving " \
+								   "queries. Alternatively, try increasing the value of pg_stat_monitor.pgsm_max.")));
+				disable_error_capture = false;
```
artemgavrilov, line 2050: Not for now but just an idea, it looks like instead of manually setting `disable_error_capture` variable we can write another function to pass to ereport. Like `ereport(WARNING, (errcode(ABC), errmsg("abc"), errnorecord(), ...`

In `errnorecord` set `disable_error_capture` and in error hook always reset it.

jeltz: Since there is only one such call I do not think it is worth it but great idea! :)

artemgavrilov (approved):

PR #538, opened 2025-06-03
PG-1621: fix cmd_type mostly showing 0 values

dutow:
This was actually caused by two bugs internally:

* cmd_type was only set in some codepaths, other parts of the code never set a value. Depending on which query / how was executed, it was possibly never changed (after a reset to 0)
* the update first set the cmd_type, then reset all counters. As the cmd_type is stored within the counters for some reason, this reset its value to 0 in most execution paths, even if it was corretly set before.

And according to this the fix is simple:

* cmd_type is now set in all codepaths except for failing queries, as we only have the error string in this case, without the type.
* in the update logic, we again overwrite cmd_type with the proper value after a reset


dutow:
Remaining issue: postgres commit 7103ebb7aae8 added the `MERGE` command type, which is missing from the list in pgsm. The pgsm functions currently assume that command type 6 is `NOTHING`, which is true for postgres 14 and before, but is incorrect for later.

Another problem is that `MERGE` wasn't added to the end, but before `UTILITY`, that's why it causes a test failure here, I'll have to add a condition based on the postgres version to the `get_cmd_type` function. We will also have to increment the version number / add a new sql file to fix this.

dAdAbird (approved):

pg_stat_monitor--2.1--2.2.sql
```diff
@@ -0,0 +1,22 @@
+/* contrib/pg_stat_monitor/pg_stat_monitor--2.0--2.1.sql */
```
artemgavrilov, line 1:
```suggestion
/* contrib/pg_stat_monitor/pg_stat_monitor--2.1--2.2.sql */
```

artemgavrilov (approved):

pg_stat_monitor--2.1--2.2.sql
```diff
@@ -0,0 +1,22 @@
...
+
+-- complain if script is sourced in psql, rather than via CREATE EXTENSION
+\echo Use "ALTER EXTENSION pg_stat_monitor" to load this file. \quit
+
+CREATE OR REPLACE FUNCTION get_cmd_type (cmd_type INTEGER) RETURNS TEXT AS
+$$
+SELECT
+    CASE
+        WHEN cmd_type = 0 THEN ''
```
artemgavrilov, line 10: BTW it looks like this line is redundant. For any value that it out of range this function also returns nothing.

PR #542, opened 2025-06-18
Fix comment parsing logic at two places

dutow:
* During query normalization, PGSM used a custom loop to remove comments, but the logic was incorrect. It removed the starting characters of the comment, and left everything else there.
* During comment extraction, PGSM used a very slow regex that caused significant slowdowns. This commit simply removes the regex dependency and uses the same (corrected) loop from the normalization logic.


pg_stat_monitor.c
```diff
@@ -2715,12 +2700,19 @@ get_pgsm_query_id_hash(const char *norm_query, int norm_len)
 		 */
 		if (*norm_q_iter == '/' && *(norm_q_iter + 1) == '*')
 		{
-			while (*norm_q_iter && *norm_q_iter != '*' && *(norm_q_iter + 1) != '/')
+			while (*norm_q_iter && *(norm_q_iter + 1) && (*norm_q_iter != '*' || *(norm_q_iter + 1) != '/'))
```
jeltz, line 2703: Will anything break when queries no longer hash to the same thing due to correcting this? I suspect it is fine but I am not familiar enough with PSM to know,

pg_stat_monitor.c
```diff
@@ -3970,46 +3962,91 @@ get_histogram_timings(PG_FUNCTION_ARGS)
...
-	while (total_len < max_len)
-	{
-		rc = regexec(&preg_query_comments, s, nmatch, &pmatch, 0);
-		if (rc != 0)
-			break;
+	size_t		curr_len = 0;
+	const char *q_iter = query;
 
-		comment_len = pmatch.rm_eo - pmatch.rm_so;
+	if (!pgsm_extract_comments)
```
jeltz, line 3984: Maybe I failed to read it but this wasn't mentioned in the commit message, right?

pg_stat_monitor.c
```diff
@@ -3970,46 +3962,91 @@ get_histogram_timings(PG_FUNCTION_ARGS)
...
+			}
+		}
 
-		/* Not 1st iteration, append ", " before next comment. */
-		if (s != query)
+		/*
+		 * Skip single line comments, + 1 is safe even if we've reach end of
+		 * string
+		 */
+		if (*q_iter == '-' && *(q_iter + 1) == '-')
```
jeltz, line 4029: Does this really belong in a bugfix? This changes the behvior to also parse out single line comments.

pg_stat_monitor.c
```diff
@@ -3970,46 +3962,91 @@ get_histogram_timings(PG_FUNCTION_ARGS)
...
+				if (!append_comment_char(comments, max_len, &curr_len, *q_iter))
+					return;
+				q_iter++;
+			}
+		}
 
-		/* Not 1st iteration, append ", " before next comment. */
-		if (s != query)
+		/*
+		 * Skip single line comments, + 1 is safe even if we've reach end of
```
jeltz, line 4026: Looks to me like we handle them rather than skipping them here.

jeltz (changes requested):
A minor thing is that I think this commit would make sense to split into multiple for an easier to read history.

pg_stat_monitor.c
jeltz: Also if we add this there should be a test case.

dutow: Good question, based on the name of the feature, I assumed that it should be included, but I can remove it if you think that makes more sense.

We barely have any testcases for comments at all, only 1 query with a single comment. We have to increase test coverage, but I wanted to delay that to the next ticket with the remaining improvements.

pg_stat_monitor.c
dutow: You are right, I'll include it.

PR #592, opened 2025-12-05
Fix buffer overflow issue

dutow:
## Correct the size of tmp array

The snprintf using it should work with any int, which means this string
has to fit ",-2147483648\0", which is 12 characters + a null terminator.

Otherwise it could lead to a buffer overflow, corrupting other variables
on the stack, and maybe causing a crash too?

## Fix resp_calls array size

This got broken during a refactoring in https://github.com/percona/pg_stat_monitor/commit/de712829223e0aefed62a1bbfe206694fa15993e. We need the plus 2
items because of the overflow/underflow bucket, without that we can
possibly overindex this array and corrupt other counters.

In earlier versions we even had the encoding variable stored directly
after this array, which possibly caused crashes or incorrectly working
string functions.

### DIFF
--- a/pg_stat_monitor.h
+++ b/pg_stat_monitor.h
@@ -325,8 +325,9 @@ typedef struct Counters
 	JitInfo		jitinfo;
 	ErrorInfo	error;
 	Wal_Usage	walusage;
-	int			resp_calls[MAX_RESPONSE_BUCKET];	/* execution time's in
-													 * msec */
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
AndersAstrand: strcat here, however, can lead to a buffer overflow if len is too large. it should probably be fixed even if len is known to always be small.

AndersAstrand (approved):

PR #616, opened 2026-02-10
Trust semver in github actions

AndersAstrand:
We don't use github actions for anything production critical, so it seems fine to trust that the promises of semver is kept for these actions rather than having dependabot spam us with patch updates.


jeltz (approved):
Since we do not use it to build any artifacts and only for our CI I do not think pinning the versions is worth the hassle.

PR #634, opened 2026-03-19
PG-2229 Memory leaks once query executed via SPI interface

artemgavrilov:
PG-2229

### Description

Before attached our memory context cleanup callback to `MessageContext`, however this context doesn't exist when query executed via SPI. That caused memory leak. In this PR we explicitly reset our context when we no longer need its data, as there is no other context that matches our lifetime requirements.

### How to test

I used `worker_spi` extension from `postgres/src/test/modules/worker_spi`

1. Start postgresql with PGSM added to preload libs
2. `CREATE EXTENSION worker_spi;`
3. `SELECT worker_spi_launch(1) IS NOT NULL;`
4. ` SELECT pid, query FROM pg_stat_activity;` find PID of the process that executes query `WITH deleted AS (DELETE FROM...`
5. `SELECT pg_log_backend_memory_contexts(<PID>);`
6. Check postgres logs for such record `LOG:  level: 2; pg_stat_monitor local store: 8192 total in 1 blocks; 7952 free (0 chunks); 240 used`.

Without this patch used space will constantly grow over time, while with the patch it will stay constant.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -1625,6 +1625,7 @@ pgsm_delete_entry(uint64 queryid)
 		pfree(entry->query_text.query_pointer);
 		entry->query_text.query_pointer = NULL;
 		lentries = list_delete_last(lentries);
+		pfree(entry);
 		return;
 	}
 
@@ -1642,6 +1643,7 @@ pgsm_delete_entry(uint64 queryid)
 			pfree(entry->query_text.query_pointer);
 			entry->query_text.query_pointer = NULL;
 			lentries = list_delete_cell(lentries, lc);
+			pfree(entry);
 			return;
 		}
 	}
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -1013,175 +1000,192 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
-		PG_CATCH();
+		pgsm_delete_entry(pstmt->queryId);
+	}
+	PG_FINALLY();
+	{
+#if PG_VERSION_NUM >= 170000
+		if (nesting_level == 0)
+#else
+		if ((nesting_level + plan_nested_level) == 0)
+#endif
```
artemgavrilov: Here we have to use try/finally because our context created under TopMemoryContext, so it lives forever. In this hook we may have an exception and we want to cleanup memory in this case too.

jeltz:
Given the new idea for design for the memory I think we can close this.

PR #635, opened 2026-03-24
Refactoring

artemgavrilov:
### Description

Some refactoring:

1. Remove redundant argument from pgsm_create_hash_entry function
2. Move declarations that used in one place to that place (in the same file)
3. Little cleanup of main header file

### DIFF
--- a/hash_query.c
+++ b/hash_query.c
@@ -19,7 +19,19 @@
 
 #include "pg_stat_monitor.h"
 
+typedef struct pgsmLocalState
+{
+	pgsmSharedState *shared_pgsmState;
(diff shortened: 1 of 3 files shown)


AndersAstrand (changes requested):
sence->sense in the first commit message.

I think the second commit is an improvement, but I think we should introduce guc.h and hash_query.h to make things even better than having those "sections" in the one h-file.

If you don't feel like doing that we should commit this as is though.

jeltz (approved):
Pretty sure there is more stuff in the `pg_stat_monitor.h` which could be moved into `pg_stat_monitor.c` but this is an iprovement on its own.

PR #642, opened 2026-04-02
Set unit for pg_stat_monitor.pgsm_query_max_len to bytes

jeltz:
Given the previous confusion we had in our documentation about characters vs bytes I think adding a unit would make it clear that the limit is measured in bytes.


AndersAstrand (approved):
I think `INT_MAX` is probably not the correct max value here, (maybe `MaxAllocSize`?) but this PR is still good for documenting the intended unit.

jeltz:
Agreed, but I tried not to fix all problems I saw in PGSM. Then I would never have finished.

PR #648, opened 2026-04-08
Make sure to load the upgrade script for the future 2.4

jeltz:
We added an upgrade script for the future 2.4 release in commit 3481ee75eba8e333d402e2bf2511823f950eefee but we never actually bumped the version in the control file so the file was never loaded and therefore the tests were never updated to match dropping the `pgsm_create_13_view()` function.

## Questions

- Should we make sure the upgrade script for the upcoming release is ran? I think so because otherwise we may have broken code in the repo.
- What should the upcoming unreleased version be named? 2.4, 2.4-dev, dev, something else?


AndersAstrand:
Here is one idea:

If there is ongoing SQL changes in `main`, we change the version in the control file to "development" and make upgrade scripts and whatnot for that version. This will then have to be renamed and such in the release branch.

This seems less prone to mistakes to me than trying to predict what the next release version will be named.

jeltz:
Yeah, my idea was basically call the next version for "development" or "dev".

artemgavrilov (approved):

AndersAstrand (approved):

PR #654, opened 2026-05-14
Add PostgreSQL 19 compatibility

adeshthack:
Description:

  PG19 introduces several API changes that break the build. This PR adds

  C code changes (pg_stat_monitor.c, hash_query.c, pg_stat_monitor.h):
  - Adapt to planner_hook gaining ExplainState *es parameter
  - Handle QueryDesc->totaltime rename to QueryDesc->query_instr with
    executor-managed instrumentation lifecycle (query_instr_options)
  - Fix LWLockNewTrancheId() now requiring a name argument
  - Fix ShmemInitHash() dropping max_size parameter (5->4 args)
  - Handle const JumbleState* in post_parse_analyze_hook
  - Remove escape_string_warning (dropped from core_yy_extra_type)
  - Add missing #include headers guarded by PG_VERSION_NUM >= 190000

  Test changes:
  - Add regression/expected/level_tracking_3.out for PG19
  - Add t/expected/007_settings_pgsm_query_shared_buffer.out.19
  - Add PG19 entry to t/018_column_names.pl

  All changes use #if PG_VERSION_NUM >= 190000 / #else / #endif guards.
  Tested against PG 19devel — all 22 regression tests and 34 TAP tests pass.

  Links:
  Related PostgreSQL 19 commits:
  - https://github.com/postgres/postgres/commit/b9e34283 (planner_hook ExplainState)
  - https://github.com/postgres/postgres/commit/5765fbd5 (QueryDesc->query_instr)

  Upstream CI only tests PG 14-18:
  - https://github.com/percona/pg_stat_monitor/blob/main/.github/workflows/matrix.yml

### DIFF
--- a/hash_query.c
+++ b/hash_query.c
@@ -122,7 +122,11 @@ pgsm_startup(void)
 		pgsm->raw_dsa_area = p;
 		dsa = dsa_create_in_place(pgsm->raw_dsa_area,
 								  pgsm_query_area_size(),
+#if PG_VERSION_NUM >= 190000
+								  LWLockNewTrancheId("pg_stat_monitor_dsa"), 0);
+#else
 								  LWLockNewTrancheId(), 0);
+#endif
 		dsa_pin(dsa);
 		dsa_set_size_limit(dsa, pgsm_query_area_size());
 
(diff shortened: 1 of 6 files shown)


lfittl:
To add to this, I'd encourage the extension authors to also look at this commit https://git.postgresql.org/gitweb/?p=postgresql.git;a=commitdiff;h=49cc0d41488b92a1d547e278b459c3b224393c8f;hp=51098839cfb40dc496bb48fe4ea14d19170099ac

That commit made `JumbleState` const in Postgres 19. As part of testing that commit, I've verified `pg_stat_monitor` can be adapted by making a change like this: https://github.com/lfittl/pg_stat_monitor/commit/054f347344a380f7a59cb444685db71301669c61

(I also had a separate commit for other PG19 fixes, which is very similar to this PR, FWIW: https://github.com/lfittl/pg_stat_monitor/commit/77f6ec1b7c932530f32208059ff6c020bb44a574)

jeltz:
Thanks for the pull request and sorry for the late response, but I think adding PG 19 support is a bit more than this. For example I would want some tests to make sure for example PGQ is supported. And I also wonder if we should not move to the new API for hashes in shared memory.

jeltz:
Closing this in favour of #806.

PR #656, opened 2026-05-15
Use TopTransactionContext lifetime for PGSM local data

artemgavrilov:
PG-0

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -391,13 +391,9 @@ pgsm_post_parse_analyze_internal(ParseState *pstate, Query *query, JumbleState *
 
 	if (callback_setup == false)
 	{
-		/*
-		 * If MessageContext is valid setup a callback to cleanup our local
-		 * stats list when the MessagContext gets reset
-		 */
-		if (MemoryContextIsValid(MessageContext))
+		if (MemoryContextIsValid(TopTransactionContext))
 		{
-			MemoryContextRegisterResetCallback(MessageContext, &mem_cxt_reset_callback);
+			MemoryContextRegisterResetCallback(TopTransactionContext, &mem_cxt_reset_callback);
 			callback_setup = true;
 		}
 	}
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -391,13 +391,9 @@ pgsm_post_parse_analyze_internal(ParseState *pstate, Query *query, JumbleState *
...
 	{
-		/*
-		 * If MessageContext is valid setup a callback to cleanup our local
-		 * stats list when the MessagContext gets reset
-		 */
-		if (MemoryContextIsValid(MessageContext))
+		if (MemoryContextIsValid(TopTransactionContext))
 		{
-			MemoryContextRegisterResetCallback(MessageContext, &mem_cxt_reset_callback);
+			MemoryContextRegisterResetCallback(TopTransactionContext, &mem_cxt_reset_callback);
```
artemgavrilov, line 396: It seems that TopTransactionContext always exists, no matter how query was executed.

pg_stat_monitor.c
```diff
@@ -1029,11 +1025,32 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 		BufferUsage bufusage_start = pgBufferUsage;
 		WalUsage	walusage;
 		WalUsage	walusage_start = pgWalUsage;
-		pgsmEntry  *entry = pgsm_create_hash_entry(0, queryId, NULL);
```
artemgavrilov, line 1032: We cannot invoke this function here anymore, as some utility queries create and finish multiple transactions during its processing. Vacuum for example. But we still need username and application name that triggered the query. Without this `SET application_name = new app` will log `new app` as query tirgger.

pg_stat_monitor.c
```diff
@@ -1710,6 +1746,9 @@ pgsm_create_hash_entry(uint64 bucket_id, int64 queryid, PlanInfo *plan_info)
 
 	if (pgsm_track_application_names)
 	{
+		/* TODO: Move app name fetching out of this function
+		 * because at this point app name may be changed by the query itself
+
 		/* Get the application name and set appid */
 		app_name_len = pg_get_application_name(app_name, APPLICATIONNAME_LEN);
```
artemgavrilov, line 1753: This line updates global variable `app_name`, we shouldn't do it here as it affects what app name will be logged for `SET application_name = new app` query.

pg_stat_monitor.c
```diff
@@ -1029,11 +1025,32 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 		BufferUsage bufusage_start = pgBufferUsage;
 		WalUsage	walusage;
 		WalUsage	walusage_start = pgWalUsage;
-		pgsmEntry  *entry = pgsm_create_hash_entry(0, queryId, NULL);
+		int64 appid;
```
dutow, line 1028: isn't an = 0 missing from here? otherwise it will remain uninitialized in the !pgsm_track_application_names case

artemgavrilov: Yep, it is. This PR was just brief draft to present the idea, it doesn't worth to review it I think.

jeltz:
Given the new idea for design for the memory I think we can close this.

Issue #651, opened 2026-05-08
SIGSEGV in pgsm_ExecutorEnd at pg_stat_monitor.c:781 during PortalCleanup (PG 18.3, correlates with autovacuum)


jameslupolthrd:
### Description

PostgreSQL 18.3 backends terminate with `SIGSEGV` in `pgsm_ExecutorEnd` at `pg_stat_monitor.c:781` during commit-time `PortalCleanup`. Crashes correlate tightly with autovacuum completion on the relation the failing query targets: the postmaster logs an autovacuum-done line for a table, and within 1–2 seconds a backend running a query against that same table segfaults. The postmaster then terminates all other backends, replays WAL, and the database returns in ~1 second.

This is reproducible across four independent RDS PostgreSQL 18.3 instances sharing the same `shared_preload_libraries` setting. A control instance in the same account with pg_stat_monitor absent from `shared_preload_libraries` has zero crashes over the same window. Removing `pg_stat_monitor` from `shared_preload_libraries` is a confirmed workaround — all four instances have been crash-free since the extension was removed on 2026-04-30.

> **Note:** This report was prepared with AI assistance. Some information is inferred rather than directly observable because the affected instances run on AWS RDS (a managed service), which does not expose the unsanitized stack trace, the exact binary build, or `/rdsdbdata/log/error/` core dumps to customers. AWS Support extracted the sanitized stack trace below on our behalf. We have provided everything we were able to obtain.

### Expected Results

`pgsm_ExecutorEnd` should safely no-op or skip its stats update when the executor state it relies on has been invalidated during portal cleanup, rather than dereferencing a NULL/freed pointer and crashing the backend.

#### Sanitized stack trace (from AWS Support, crash on 2026-04-30 16:56:58 UTC, backend PID 12539)

```
#0  pgsm_ExecutorEnd (queryDesc=...) at pg_stat_monitor.c:781
#1  PortalCleanup (portal=...) at portalcmds.c:309
#2  PortalDrop (portal=..., isTopCommit=true) at portalmem.c:502
#3  PreCommit_Portals (isPrepare=false) at portalmem.c:756
#4  CommitTransaction () at xact.c:2269
#5  CommitTransactionCommandInternal () at xact.c:3202
#6  CommitTransactionCommand () at xact.c:3165
#7  finish_xact_command () at postgres.c:2832
#8  PostgresMain (...) at postgres.c:4969
#9  BackendMain (...) at backend_startup.c:129
[...]

PR #658, opened 2026-05-25
PG-2395 Fetch datname and username from syscache in pgsm_store()

AndersAstrand:
Resolve them from the catalog in pgsm_store() rather than in pgsm_create_hash_entry(). We have a bug report we have not been able to reproduce where queryDesc->totaltime is freed mid-call in pgsm_ExecutorEnd(), most likely by cache invalidation processing during these syscache lookups when they ran inside pgsm_create_hash_entry(). The misbehaving callback is most likely from another extension or the RDS fork of PostgreSQL, since standard invalidation callbacks does not seem to touch the relevant memory context.

I didn't change the undefined behavior of passing NULL to snprintf() in this PR, but we should fix it in the future (i believe glibc inserts a placeholder like `(null)` for us which is the existing behavior before this PR.)

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -1695,8 +1695,6 @@ pgsm_create_hash_entry(uint64 bucket_id, int64 queryid, PlanInfo *plan_info)
 	int			sec_ctx;
 	bool		found_client_addr = false;
 	MemoryContext oldctx;
-	char	   *datname = NULL;
-	char	   *username = NULL;
 
 	/* Create an entry in the pgsm memory context */
 	oldctx = MemoryContextSwitchTo(GetPgsmMemoryContext());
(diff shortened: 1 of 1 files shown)


jeltz (changes requested):
I don't feel the commit message actually explains why this fix makes sense. Is it because it is now allocated in another memory context? If so that was not clear from the commit message.

AndersAstrand:
> I don't feel the commit message actually explains why this fix makes sense. Is it because it is now allocated in another memory context? If so that was not clear from the commit message.

That's great feedback! It's hard to write with the right level of detail when you have all the context in your head that the reader probably doesn't :). Here is a full explanation of my reasoning. Do you think all of this should go in the commit message?

The bug report #651 includes a stack trace that shows that by the time we read `queryDesc->totaltime`, the pointer isn't valid.

A lot of my analysis was based on what we do differently compared to `pg_stat_statements` which our use of `queryDesc->totaltime` is based on, and nothing related to that pointer specifically differed. We allocate and set the pointer at the same time in the same way in the same relation to hook chaining. And we read the pointer at the same time in the same way and with the same relation to hook chaining.

Now the issue _could_ also be present in `pg_stat_statements` and the crash would happen there if the two were swapped in the `shared_preload_libraries`, but this felt unlikely to me.

One of the things we do, that `pg_stat_statements` does not, is doing syscache lookups before we read the contents of `queryDesc->totaltime`. And since the reporter of the bug saw a correlation with aggressive autovacuum I started to suspect that maybe the callback chains triggered by the syscache lookups processing the invalidation messages produced by autovacuum could have something to do with it. Possibly because some other extension either ruined the memory context we had allocated totaltime in, or maybe overwrote the pointer with something that was just garbage after they were done with it.

I had no way of verifying this suspicion however since I was unable to reproduce the bug, so I simply moved the syscache lookups to somewhere after we're done with `queryDesc->totaltime`. `pgsm_store()` seemed reasonable.

jeltz:
But if the issue is `totaltime` why would you think changing how we work with `datname` and `username` would help?

jeltz:
Close this in favour of #811.

PR #662, opened 2026-06-04
Use PG_MODULE_MAGIC_EXT macro in PostgreSQL 18

jeltz:
PostgreSQL 18 introduced the very useful `PG_MODULE_MAGIC` macro which makes it possible to see which exact version of the library is loaded into PostgreSQL.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -64,9 +64,13 @@ typedef enum pgsmVersion
 	PGSM_V2_3,
 } pgsmVersion;
 
-PG_MODULE_MAGIC;
-
 #define BUILD_VERSION                   "2.4.0"
+
+#if PG_VERSION_NUM >= 180000
+PG_MODULE_MAGIC_EXT(.name = "pg_stat_monitor",.version = BUILD_VERSION);
+#else
+PG_MODULE_MAGIC;
+#endif
 
 /* Number of output arguments (columns) for various API versions */
 #define PG_STAT_MONITOR_COLS_V1_0    52


pg_stat_monitor.c
```diff
@@ -64,10 +64,14 @@ typedef enum pgsmVersion
 	PGSM_V2_3,
 } pgsmVersion;
 
-PG_MODULE_MAGIC;
-
 #define BUILD_VERSION                   "2.4.0"
 
+#if PG_VERSION_NUM >= 180000
```
artemgavrilov, line 69: Also can be done with `#ifdef`, but not sure if it'll better

pg_stat_monitor.c
```diff
@@ -64,10 +64,14 @@ typedef enum pgsmVersion
 	PGSM_V2_3,
 } pgsmVersion;
 
-PG_MODULE_MAGIC;
-
 #define BUILD_VERSION                   "2.4.0"
```
artemgavrilov, line 67: I wonder if we should have version only in one place. Here is for example how PGXN recommends to do it: https://manager.pgxn.org/howto#neworder

i.e. keep version in META.json

artemgavrilov (approved):

pg_stat_monitor.c
jeltz: Yeah, we probably should. Or at least add a test case for it to make sure we do not diverge.

pg_stat_monitor.c
jeltz: I don't think it is better because then it will be hard to remove the old code when we stop supporting PG 17. I know: thinking very far ahead!

PR #664, opened 2026-06-05
Fix and refactor get_histogram_timings

artemgavrilov:
Use of StringInfoData type to build resulting string fixes overflow for big amount of buckets. Now output is valid for any buckets amount.

PG-0

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -2964,8 +2964,6 @@ comp_location(const void *a, const void *b)
 		return 0;
 }
 
-#define MAX_STRING_LEN	1024
-
 /* Convert array of integers into Text datum */
 static Datum
 intarray_get_datum(int32 arr[], int len)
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -3181,34 +3181,34 @@ get_histogram_timings(PG_FUNCTION_ARGS)
 {
 	double		b_start;
 	double		b_end;
```
jeltz: When you are already touching this code maybe these two should be moved inside the loop?

jeltz (approved):
Looks great but had one small improvement you can optionally do.

pg_stat_monitor.c
```diff
@@ -3181,34 +3181,34 @@ get_histogram_timings(PG_FUNCTION_ARGS)
...
-	int			b_count = hist_bucket_count_total;
-	int			index = 0;
-	char	   *tmp_str = palloc0(MAX_STRING_LEN);
-	char	   *text_str = palloc0(MAX_STRING_LEN);
```
jeltz, line 3187: I do not think anything uses this define anymore after this.

pg_stat_monitor.c
artemgavrilov: Done

pg_stat_monitor.c
artemgavrilov: Ah, forgot to add it to commit.
Removed.

PR #665, opened 2026-06-06
Remove two unused defines

jeltz:
These defines were not used anywhere in our code.

### DIFF
--- a/pg_stat_monitor.h
+++ b/pg_stat_monitor.h
@@ -31,17 +31,14 @@
 #define USAGE_EXEC(duration)	(1.0)
 #define USAGE_INIT				(1.0)	/* including initial planning */
 
-#define JUMBLE_SIZE				1024	/* query serialization buffer size */
-
 #define HISTOGRAM_MAX_TIME		50000000
 #define MAX_RESPONSE_BUCKET 50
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.h
```diff
@@ -31,17 +31,14 @@
...
 #define HISTOGRAM_MAX_TIME		50000000
 #define MAX_RESPONSE_BUCKET 50
 #define INVALID_BUCKET_ID	-1
 #define ERROR_MESSAGE_LEN	100
-#define REL_TYPENAME_LEN	64
 #define REL_LST				10
-#define REL_LEN				132 /* REL_TYPENAME_LEN * 2 (relname + schema) + 1
-								 * (for view indication) + 1 and dot and
-								 * string terminator */
+#define REL_LEN				132 /* NAMEDATALEN * 2 (relname + schema) + 1
```
dutow: can we make this an actual expression instead of explaining a magic number in the comment after it?

jeltz: Yeah, probably a good thing but the interesting thing is that if we do so we will end up with 131 and not 132.

jeltz: Fixed.

artemgavrilov (approved):

PR #666, opened 2026-06-06
Use anonymous struct for histograms instead of array

jeltz:
The intent is much clearer with an anonymous struct rather than an array with two entries and an enum to index it.

@artemgavrilov what do you think of this? Should I have used a named struct instead?


artemgavrilov (approved):
I like it! 

 anonymous struct probably fine

PR #671, opened 2026-06-08
Rewrite time_diff function

artemgavrilov:
Change the fucntion the way it first does subtraction and then converts result to milliseconds to keep precision.

Or maybe it's worth convert it to macro @jeltz ?


jeltz (approved):
Looks good and I think we should try to stop writing macros and use more functions instead.

PR #675, opened 2026-06-09
Use StringInfo when building realtions list

jeltz:
Instead of doing our own string building with `snprintf()` we should just use PostgreSQL's `StringInfoData` type.

### DIFF
--- a/pg_stat_monitor.c
+++ b/pg_stat_monitor.c
@@ -2309,28 +2309,19 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 		/* relations at column number 14 */
 		if (tmp.info.num_relations > 0)
 		{
-			int			j;
-			char	   *text_str = palloc0(TOTAL_RELS_LENGTH);
-			char	   *tmp_str = palloc0(TOTAL_RELS_LENGTH);
-			bool		first = true;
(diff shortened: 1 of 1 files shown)


pg_stat_monitor.c
```diff
@@ -2320,28 +2320,18 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
...
-				snprintf(tmp_str, 1024, "%s,%s", text_str, tmp.info.relations[j]);
-				snprintf(text_str, 1024, "%s", tmp_str);
+				if (j > 0)
+					appendStringInfoChar(&buf, ',');
+				appendStringInfoString(&buf, tmp.info.relations[j]);
 			}
-			pfree(tmp_str);
-			values[i++] = CStringGetTextDatum(text_str);
+
+			values[i++] = CStringGetTextDatum(buf.data);
```
artemgavrilov, line 2323: As this happens in the loop, I think it's worth to free buf.data manually

jeltz: Good point. This was a pre-existing bug, but I may as well fix it.

jeltz: Fixed!

artemgavrilov (approved):

PR #684, opened 2026-06-09
Cleanup everyting related to dshash

artemgavrilov:
dshash implementation wasn't finished and all this was just a deadcode.

### DIFF
--- a/src/hash_query.h
+++ b/src/hash_query.h
@@ -19,7 +19,6 @@
 #include <postgres.h>
 
 #include <executor/instrument.h>
-#include <lib/dshash.h>
 #include <nodes/nodes.h>
 #include <storage/lwlock.h>
 #include <storage/spin.h>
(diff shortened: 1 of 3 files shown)


src/hash_query.h
```diff
@@ -43,37 +42,6 @@
...
- * automatically get released by error.
- * But throwing an error from pg_stat_monitor would mean erroring out the user query,
- * which is not acceptable for any stat collector extension.
- *
- * Moreover, some of the pg_stat_monitor functions perform the sequence scan on the
- * hash table, while the sequence scan support for dshash table is only available
- * for PG 15 and onwards.
- * So until we figure out the way to release the locks acquired internally by dshash
- * in case of an error while ignoring the error at the same time, we will keep using
- * the classic shared memory hash table.
```
artemgavrilov, line 65: So yes, it's wasn't finished.

jeltz (changes requested):
Looks good except that we can also remove a `PG_TRY()`, right?

src/hash_query.h
jeltz: Yeah, that sounds very broken.

artemgavrilov:
> Looks good except that we can also remove a `PG_TRY()`, right?

Absolutely! Missed this thing 🤦‍♂️

Updated.

src/pg_stat_monitor.c
```diff
@@ -1898,24 +1898,13 @@ pgsm_store(pgsmEntry *entry)
...
-			PG_RE_THROW();
-		}
-		PG_END_TRY();
+		shared_hash_entry = hash_entry_alloc(pgsm, &entry->key);
 
 		if (shared_hash_entry == NULL)
 		{
 			pgsm_lock_release(pgsm);
 
-			if (DsaPointerIsValid(dsa_query_pointer))
```
jeltz, line 1917: Removing this is probably good but I do not think it belongs in the same PR or at least not the same commit.

jeltz (changes requested):

src/pg_stat_monitor.c
artemgavrilov: Extracted this change into separate commit

jeltz (approved):

PR #687, opened 2026-06-10
Use PostgreSQL::Version to parse PostgreSQL version in TAP tests

jeltz:
`PostgreSQL::Version` was introduced in PostgreSQL 14 but even before that we should probably have used Perl to parse the version string rather than using `sed` and `awk`.

PG-0


jeltz:
Ah, seems the situation with PG 14 is a bit more complicated. Apparently the module was called `PostgresVersion` back then.

PR #694, opened 2026-06-11
Simplify logic for initialization check and OOM

jeltz:
By setting the flag for successful initialization in our shared memory startup hook we can simplify the logic for the initialization check. And while working in this area we also simplify the OOM code a bit.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -344,8 +344,6 @@ _PG_init(void)
 
 	nested_queryids = (int64 *) malloc(sizeof(int64) * max_stack_depth);
 	nested_query_txts = (char **) calloc(max_stack_depth, sizeof(char *));
-
-	system_init = true;
 }
 
 /*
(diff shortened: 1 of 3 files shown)


src/pg_stat_monitor.c
```diff
@@ -2975,7 +2975,7 @@ pgsm_emit_log_hook(ErrorData *edata)
 	if (MyProc == NULL)
 		goto exit;
 
-	if (edata->elevel >= WARNING && !disable_error_capture && IsSystemOOM() == false)
+	if (edata->elevel >= WARNING && !disable_error_capture && !IsSystemOOM())
```
artemgavrilov, line 2996: It seems we can use `pgsm_oom` here as well as we have `IsSystemInitialized` guard above, no?

jeltz: Maybe. Maybe we should jsut remove the helper fucntion? I will think about this.

artemgavrilov (approved):

PR #698, opened 2026-06-14
Wrap pg_regress in a TAP test to setup cluster automatically

jeltz:
This makes the pg_regress tests much easier to run given that they then can run in a clean cluster and nobody has to start and stop a cluster manually.

For some reason our Meson support did not run any TAP tests which means the tests are no longer runnable from Meson. This should likely be fixed as a separate pull request.

PG-0


t/000_regress.pl
```diff
@@ -0,0 +1,67 @@
...
+	'--port' => $node->port,
+	'--dbname' => 'pgsm_regression',
+	'--inputdir' => ($ENV{'top_builddir'} || '.') . '/regression',
+	'--outputdir' => ($ENV{'top_builddir'} || '.') . '/regression',
+	%{ $PGSM::PG_MAJOR_VERSION >= 16 ? { '--expecteddir' => ($ENV{'top_builddir'} || '.') . '/regression' } : {} },
+	@tests,
+],
+'1>' => $orig_stdout,
+'2>' => $orig_stderr
+or exit $?;
```
dutow, line 65: Is this correct? IPC::Run::run shifts the exit code by 8, and exit takes modulo 256.

jeltz:
Not sure this is worth doing since the TAP protocol output was first added in PG16.

PR #713, opened 2026-06-19
PG-2122 Fix broken client IP code

### LINKED ISSUE #519: client_ip is always 127.0.0.1
### Description

`pg_stat_monitor.client_ip` is always 127.0.0.1, even though `pg_stat_activity.client_addr`  has correct IPs that are not 127.0.0.1

### Expected Results

See the same IP in `pg_stat_monitor.client_ip` and `pg_stat_activity.client_addr`

### Actual Results

All values for `client_ip` are 127.0.0.1

### Version

- PostgreSQL 17.2 - Percona Server for PostgreSQL 17.2.1 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 14.2.1 20240801 (Red Hat 14.2.1-1), 64-bit
- pg_stat_monitor 2.1

### Steps to reproduce

1. Install the extention:
    ```
    git clone https://github.com/percona/pg_stat_monitor.git /tmp/pg_stat_monitor && \
    cd /tmp/pg_stat_monitor && \
    make USE_PGXS=1 && make USE_PGXS=1 install && \
    cd && \
    rm -rf /tmp/pg_stat_monitor
    ```
2. Apply extention              
    ```
    ALTER SYSTEM SET shared_preload_libraries = 'pg_stat_monitor';
    /usr/pgsql-17/bin/pg_ctl reload
    /usr/pgsql-17/bin/pg_ctl restart
    CREATE EXTENSION pg_stat_monitor;
    ```
3. verify client IP is not 127.0.0.1:
    ```sql
    SELECT DISTINCT client_addr FROM pg_stat_activity;
    ```
4. Run any query.
5. The query should appear in `pg_stat_monitor` with the correct IP:
    ```sql
    SELECT DISTINCT client_ip FROM pg_stat_monitor;
    ```

### Relevant logs

```Shell

```

### Code of Conduct

- [x] I agree to follow Percona Community Code of Conduct

jeltz:
Since `pg_get_backend_status()` contained a typo where we incorrectly tried to look up the BackendStatus by PID instead of by ProcNumber we almost never could see the right IP address.

But instead of fixing this we just make our lives easier by not looking at the BackendStatus at all for things related to our own backend. That seems quite wasteful and overly complex when we can access the fields directly.

The IP address code is still dodgy with its defaulting to 127.0.0.1 on UNIX sockets and its failure to support IPv6.

This might fix #519

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -208,7 +208,7 @@ PG_FUNCTION_INFO_V1(pg_stat_monitor);
 PG_FUNCTION_INFO_V1(get_histogram_timings);
 PG_FUNCTION_INFO_V1(pg_stat_monitor_hook_stats);
 
-static uint pg_get_client_addr(bool *ok);
+static uint pg_get_client_addr(void);
 static int	pg_get_application_name(char *name, int buff_size);
 static PgBackendStatus *pg_get_backend_status(void);
 static Datum intarray_get_datum(const int32 *arr, int len);
@@ -1261,18 +1261,14 @@ pg_get_application_name(char *name, int buff_size)
 }
 
 static uint
-pg_get_client_addr(bool *ok)
+pg_get_client_addr(void)
 {
 	PgBackendStatus *beentry = pg_get_backend_status();
 	char		remote_host[NI_MAXHOST];
 	int			ret;
 
-	remote_host[0] = '\0';
-
 	if (!beentry)
 		return ntohl(inet_addr("127.0.0.1"));
-
-	*ok = true;
 
 	ret = pg_getnameinfo_all(&beentry->st_clientaddr.addr,
 							 beentry->st_clientaddr.salen,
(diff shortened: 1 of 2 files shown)


src/pg_stat_monitor.c
```diff
@@ -1206,76 +1205,33 @@ pgsm_hash_string(const char *str, int len)
...
-	{
-		PgBackendStatus *beentry;
-
-		beentry = pg_get_backend_status();
-
-		if (!beentry)
-			strlcpy(name, "unknown", buff_size);
-		else
-			strlcpy(name, beentry->st_appname, buff_size);
-	}
```
artemgavrilov, line 1257: We had a performance issue with app name. I assume there are cases when `application_name` is empty for some reason and we fall back to backend status that an expensive operation.

artemgavrilov (approved):

src/pg_stat_monitor.c
jeltz: Hahaha, that is funny. Is there a bug reported for that?

artemgavrilov: This is was the original report: https://perconadev.atlassian.net/browse/PG-644

And here we added option to disable app name tracking as a solution: https://github.com/percona/pg_stat_monitor/pull/469

jeltz: So maybe we no longer need to GUC. :rofl:

artemgavrilov: Probably) But better to benchmark it anyway before)

PR #723, opened 2026-06-19
Attempt at cleaning up getting the next bucket

jeltz:
The code for getting the next bucket was unnecessarily complex and hard to read. But is this suggestion good enough? Maybe we need to do an even bigger rewrite of the bucketing? Especially since the whole mess with two atomics and taking an exclusive lock here is pretty odd and unclear if it is even correct for example if we skip some buckets. Will they get cleared out then?

PG-0


jeltz:
We have already merged a similar version.

PR #724, opened 2026-06-19
Make sure to hold the entry lock while resetting

jeltz:
When we switch buckets we believe we may have to reset hash table entries. if that is actually the case is a bit unclear since we also have code which tries to delete all old entries when switching bucket, but we may at least clean up the code so we do not modify entries in shared memory without holding a lock.

While at it we also move the taking and releasing of spin locks to a more sensible location.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -238,7 +238,6 @@ static void pgsm_update_entry(pgsmEntry *entry,
 							  const struct JitInstrumentation *jitusage,
 							  int parallel_workers_to_launch,
 							  int parallel_workers_launched,
-							  bool reset,
 							  pgsmStoreKind kind);
 static void pgsm_store(pgsmEntry *entry);
 
@@ -741,7 +740,6 @@ pgsm_ExecutorEnd(QueryDesc *queryDesc)
 						  0,	/* parallel_workers_to_launch */
 						  0,	/* parallel_workers_launched */
 #endif
-						  false,	/* reset */
 						  PGSM_EXEC);	/* kind */
 
 		pgsm_store(entry);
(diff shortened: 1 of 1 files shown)


artemgavrilov (approved):

src/pg_stat_monitor.c
```diff
@@ -1866,6 +1840,20 @@ pgsm_store(pgsmEntry *entry)
...
+
+	/*
+	 * Start collecting data for next bucket and reset all counters and
+	 * timestamps.
+	 */
+	if (bucketid != prev_bucket_id)
+	{
+		memset(&shared_hash_entry->counters, 0, sizeof(Counters));
+		shared_hash_entry->counters.info.cmd_type = entry->counters.info.cmd_type;
+		shared_hash_entry->stats_since = GetCurrentTimestamp();
```
dutow: Is it okay to call GetCurrentTimestamp under the spinlock?

jeltz: Probably depends on your libc implementation and CPU architecture so does not seem like a great idea to call it here, hmm.

jeltz: I mean we do worse thing in this code under a spinlock but that is no excuse to add more problematic code.

jeltz: Fixed!

src/pg_stat_monitor.c
dutow, line 1233: should this function document/assert that we have to hold the lock before calling it now? (not sure if assert is possible with spinlocks, so documentation seems easier choice)

dutow (approved):

src/pg_stat_monitor.c
jeltz: Yeah, it should probably say something about it in a comment.

PR #727, opened 2026-06-20
PG-2485 Fix off-by-one bug in comment extraction

jeltz:
We could overflow the buffer by one byte due to and off-by-one bug. While fixing this bug we also rename the length variable to make the intent more obvious.

The code still does not handle multi-byte characters correctly but just not overflowing the buffer is a nice improvement.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -3075,10 +3075,13 @@ get_histogram_timings(PG_FUNCTION_ARGS)
 }
 
 static bool
-append_comment_char(char *comments, size_t max_len, size_t *idx, char c)
+append_comment_char(char *comments, size_t buf_len, size_t *idx, char c)
 {
-	if (*idx >= max_len)
+	if (*idx >= buf_len - 1)
+	{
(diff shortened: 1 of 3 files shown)


src/pg_stat_monitor.c
```diff
@@ -3107,10 +3107,13 @@ get_histogram_timings(PG_FUNCTION_ARGS)
 }
 
 static bool
-append_comment_char(char *comments, size_t max_len, size_t *idx, char c)
+append_comment_char(char *comments, size_t buf_len, size_t *idx, char c)
 {
-	if (*idx >= max_len)
+	if (*idx >= buf_len - 1)
+	{
+		comments[buf_len - 1] = 0;
```
artemgavrilov: Maybe use `\0` here, just to make it more obvious that it's string terminator?

artemgavrilov (approved):

src/pg_stat_monitor.c
jeltz: Done

PR #729, opened 2026-06-21
Remove outdated alternate pg_regress expected files

jeltz:
These files are no longer necessary on modern versions of PostgreSQL and  were accidentally left in the repo when we dropped support for those versions.

PG-0


artemgavrilov (approved):
To make tracking of alternative files easier in future maybe we should consider using PG version number instead of just sequential number? 
Like `error_15.out`, `error_17.out`

jeltz:
No, that is not possible to do.

```
/*
 * In: filename.ext, Return: filename_i.ext, where 0 <= i <= 9
 */
static char *
get_alternative_expectfile(const char *expectfile, int i)
```

artemgavrilov:
> No, that is not possible to do.
> 
> ```
> /*
>  * In: filename.ext, Return: filename_i.ext, where 0 <= i <= 9
>  */
> static char *
> get_alternative_expectfile(const char *expectfile, int i)
> ```

Hm, I bet I saw that approach somewhere. Probably there was something else then...

PR #735, opened 2026-06-30
Stop calculating current bucket or current time inside loop

jeltz:
It is expensive and can cause weird results. Better to just compare all rows in the view output with the same timestamp and the same time. This slightly changes the behavior of handling when buckets or switched mid-query but I think for the better.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -1952,18 +1952,15 @@ pg_stat_monitor(PG_FUNCTION_ARGS)
 }
 
 static bool
-IsBucketValid(uint64 bucketid)
+IsBucketValid(uint64 bucketid, TimestampTz now)
 {
 	long		secs;
 	int			microsecs;
-	TimestampTz current_tz = GetCurrentTimestamp();
 	pgsmSharedState *pgsm = pgsm_get_ss();
 
-	TimestampDifference(pgsm->bucket_start_time[bucketid], current_tz, &secs, &microsecs);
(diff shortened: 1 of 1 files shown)


src/pg_stat_monitor.c
```diff
@@ -1952,18 +1952,15 @@ pg_stat_monitor(PG_FUNCTION_ARGS)
...
-	TimestampTz current_tz = GetCurrentTimestamp();
 	pgsmSharedState *pgsm = pgsm_get_ss();
 
-	TimestampDifference(pgsm->bucket_start_time[bucketid], current_tz, &secs, &microsecs);
+	TimestampDifference(pgsm->bucket_start_time[bucketid], now, &secs, &microsecs);
 
-	if (secs > ((int64) pgsm_bucket_time * pgsm_max_buckets))
-		return false;
-	return true;
+	return secs < (int64) pgsm_bucket_time * pgsm_max_buckets;
```
dutow: Shouldn't this be `<=`? `<` is a slight behavior change.

jeltz: Maybe, not sure. But if the change is correct it deserves a comment.

dutow: I think both could work, but you didn't mention it in the commit message, so it seemed unintended.

jeltz: yes, it was. But it may be an accidental bugfix. Let me think.

jeltz: Changed to `<=` and will think on the bug as a separate PR.

dutow (approved):

artemgavrilov (approved):

PR #738, opened 2026-07-01
PG-2511 Fix buffer overflow if superuser raises max_stack_depth

jeltz:
Our code seems to make the assumption that `max_stack_depth` cannot change when allocating that arrays for nested queries, which is not the case since that GUC is SUSET.

We fix this by storing the current value of max_stack_depth when the allocations happened. An alternative would be to use buffers which grow dynamically but that would be a slightly more invasive patch so let's keep it simple.
    
Also remove a couple of redundant comments and use `palloc*()` instead of `malloc()`/`calloc()`.

PG-2511

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -269,6 +269,8 @@ static void pgsm_lock_release(pgsmSharedState *pgsm);
 void
 _PG_init(void)
 {
+	MemoryContext oldctx;
+
 	elog(DEBUG2, "[pg_stat_monitor] pg_stat_monitor: %s().", __FUNCTION__);
 
 	/*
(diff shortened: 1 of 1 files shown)


src/pg_stat_monitor.c
```diff
@@ -325,8 +328,18 @@ _PG_init(void)
 	prev_ExecutorCheckPerms_hook = ExecutorCheckPerms_hook;
 	ExecutorCheckPerms_hook = pgsm_ExecutorCheckPerms;
 
-	nested_queryids = (int64 *) malloc(sizeof(int64) * max_stack_depth);
-	nested_query_txts = (char **) calloc(max_stack_depth, sizeof(char *));
+	/*
+	 * Use max_stack_depth as a very a high and very rough estimate for
```
dutow: a very high

dutow (approved):

PR #741, opened 2026-07-01
Actually pop DDL queries from our entry stack

jeltz:
Since we for DDL set the query ID to zero we need to actually use the original query ID also when we pop from the stack, not just when we add it to the stack. This is not a memory leak since we free the whole list in `pgsm_cleanup_callback()` but it is clearly a mistake.

Also while at it change it so we only try to pop in code paths where we pushed to the stack.

PG-0


dutow (approved):
> This is not a memory leak since we free the whole list in pgsm_cleanup_callback() but it is clearly a mistake.

It does cause an assertion if a single message executes hundreds of utility statements, but that's definitely a corner case.

PR #743, opened 2026-07-01
Use INFINITY rather than -1 for upper histogram bound

jeltz:
This makes the intent of the code easier to understand.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -3008,7 +3008,7 @@ histogram_bucket_timings(int index, double *b_start, double *b_end)
 	else if (index == b_count - 1 && q_max < HISTOGRAM_MAX_TIME)
 	{
 		*b_start = q_max;
-		*b_end = -1;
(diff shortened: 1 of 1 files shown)


dutow (commented):
The comment in get_histogram_bucket is now outdated:

	/*
	 * So haven't found a histogram bucket for this query. That's only
	 * possible for the last bucket as its end time is less than 0.
	 */
	return (hist_bucket_count_total - 1);
}

jeltz:
@dutow I have fixed the comment now but I wonder if I shouldn't just change the function to just loop `hist_bucket_count_total - 2` and then return `hist_bucket_count_total - 1`. Some alterantives:

1. Loop only while `index < hist_bucket_count_total - 1`
2. Loop until we reach `q_time <= hist_bucket_timings[index].end` is true and do not check at all against `hist_bucket_count_total`
3. Add an assert before the final return telling people it is unreachable

artemgavrilov (approved):

PR #744, opened 2026-07-01
Convert histogram() to an SQL function and clean it up

jeltz:
As `histogram()` did not use any real PL/PgSQL features the intent would be clearer as an SQL function. In addition also clean it up so it is readable.

PG-0

### DIFF
--- a/pg_stat_monitor--2.3--next.sql
+++ b/pg_stat_monitor--2.3--next.sql
@@ -46,6 +46,27 @@ RETURNS text[]
 LANGUAGE sql
 RETURN string_to_array(get_histogram_timings(), ',');
 
+-- Intentionally uses a string body so we can drop and recreate the pg_stat_monitor view
+CREATE OR REPLACE FUNCTION histogram(_bucket int, _quryid int8)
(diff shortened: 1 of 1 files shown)


pg_stat_monitor--2.3--next.sql
```diff
@@ -46,6 +46,27 @@ RETURNS text[]
 LANGUAGE sql
 RETURN string_to_array(get_histogram_timings(), ',');
 
+-- Intentionally uses a string body so we can drop and recreate the pg_stat_monitor view
+CREATE OR REPLACE FUNCTION histogram(_bucket int, _quryid int8)
```
dutow, line 50: typo: _quryid

dutow (approved):

pg_stat_monitor--2.3--next.sql
jeltz: An typo I intentionally kept since this was a pure refactoring commit and I would prefer to not break the public API (the names of function argument matter since you can reference arguments by name). Maybe we should fix this but then as a separate commit or even PR.

PR #752, opened 2026-07-03
PG-2540 Fix buffer overread and simplify code for parsing comments

jeltz:
The code was hard to read for no obvious gain so this attempts to make the code read more like one would expect. Also remove a redundant comment.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -3167,13 +3167,9 @@ static bool
 append_comment_char(char *comments, size_t buf_len, size_t *idx, char c)
 {
 	if (*idx >= buf_len - 1)
-	{
-		comments[buf_len - 1] = '\0';
 		return false;
-	}
 
-	comments[*idx] = c;
-	(*idx)++;
+	comments[(*idx)++] = c;
 
 	return true;
 }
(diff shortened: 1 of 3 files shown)


dutow (commented):
There's also a possible over-read in this code. While the pr only aims for simplification, should we also fix that too, if we are already changing this area of code?

`SELECT 1 /*tail*/` or `SELECT 1 /*oops`

inner while loop consumes `*/` and points to the null. End of foor loop increases pointer again. `*q_iter` used by the for loop is now the character after null.

The unterminated comment is similar, the while loop already ends with q_iter pointing to the null byte.

jeltz:
@dutow Does this fix look correct for the over-read?

dutow (approved):
Looks good now.

We should add more coverage to this (the cases I already mentioned for asan + maybe something like `SELECT 1 /*a*//*b*/`, but that can also be a separate PR.

dutow:
Another question: since this now also includes a bugfix, it should have a proper ticket?

jeltz:
Yeah, we probably should add a test case and a proper ticket now.

jeltz:
Pushed the bugfix as a separate commit with tests and added a reference to a bug in Jira.

dutow (approved):

PR #761, opened 2026-07-05
Clean up get_pgsm_query_id_hash()

jeltz:
The function was unnecessarily convoluted so clean it up by making the parser easier to read, asserting against NULL input and wrapping the comment at the right width. Especially the code for handling whitespace was much more complicated than necessary.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -2582,11 +2582,10 @@ get_next_wbucket(pgsmSharedState *pgsm)
 }
 
 /*
- * This function expects a NORMALIZED query as the input.
- * It iterates over the normalized query skipping comments and
- * multiple spaces. All spaces are converted to ' ' so that we
- * the calculation is independent of the space type whether
- * newline, tab, or any other type. Trailing and leading spaces
+ * This function expects a NORMALIZED query as the input. It iterates over the
+ * normalized query skipping comments and multiple spaces. All spaces are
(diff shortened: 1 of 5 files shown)


dutow (commented):
`a SPACE /* */ SPACE b` now emits `a SPACE SPACE b`, previously it was `a SPACE b`, which results in a different hash. Is this something that can cause issues on upgrade? Also this means that it is no longer strictly independent of comments.

jeltz:
Ouch, good point. That was a stupid bug for me to introduce.

jeltz:
Fixed the parsing issue and added another test to detect it.

src/pg_stat_monitor.c
```diff
@@ -2595,82 +2594,71 @@ get_pgsm_query_id_hash(const char *norm_query, int norm_len)
...
-			continue;
+			while (*norm_q_iter)
+			{
+				if (*norm_q_iter == '*' && *(norm_q_iter + 1) == '/')
+				{
+					norm_q_iter++;
+					norm_q_iter++;
+					break;
+				}
+				else
```
dutow: nitpick: do we need this else? the if always breaks.

dutow (approved):

src/pg_stat_monitor.c
jeltz: Yeah, might as well remove it.

PR #766, opened 2026-07-07
Run CI actions for PGSM with assertions enabled

jeltz:
While the sanitizer checks were ran with assertions enabled I do not think that is enough. We want to test with assertions enabled on all versions of PostgreSQL that we support.

To save some CPU cycles now that we will build much more we only build with gcc against packages on Linux.
    
Also disable building with injection points for now until we have some test case which actually uses them since PostgreSQL version before 17 did not support them.

PG-0

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -19,7 +19,7 @@ jobs:
         pg_package: [psp]
         pg_version: [14, 15, 16, 17, 18]
         os: [ubuntu-24.04]
-        compiler: [gcc, clang]
+        compiler: [gcc]
         build_type: [debugoptimized]
     uses: ./.github/workflows/build-and-test.yml
     with:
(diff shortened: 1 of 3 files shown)


dutow (approved):

.github/workflows/matrix.yml
```diff
@@ -38,6 +38,25 @@ jobs:
...
+    with:
+        pg_package: ${{ matrix.pg_package }}
+        pg_version: ${{ matrix.pg_version }}
+        os: ${{ matrix.os }}
+        compiler: ${{ matrix.compiler }}
+        build_type: ${{ matrix.build_type }}
+    secrets: inherit
+
+  source:
+    name: From source
```
artemgavrilov: I would rather name it just `Source`

artemgavrilov (approved):

.github/workflows/matrix.yml
jeltz: Fixed.

PR #775, opened 2026-07-09
Various cleanup of pgsm.pm

jeltz:
- Remove special case for old PostgreSQL versions
- Change from `==` to `eq`
- Stop using sed, cut and awk when we have Perl
-  Fix formatting of a comment

PG-0


t/pgsm.pm
```diff
@@ -127,19 +126,18 @@ sub setup_files_dir
...
+	# Remove .pl from filename and store in a variable
 	my @split_arr = split /\./, $perlfilename;
 
 	my $filename_without_extension = $split_arr[0];
 
 	# Create expected filename with path
 	my $expected_filename = "${filename_without_extension}.out";
 
-	if ($PG_MAJOR_VERSION >= 15
-		and "$filename_without_extension" ==
```
artemgavrilov, line 139: Hm, we have PG 14 in our CI matrix. I wonder why it's not failed?

jeltz: Because I renamed the output file for all versions before 15 from `007_settings_pgsm_query_shared_buffer.out` to `007_settings_pgsm_query_shared_buffer.out.14`.  Now that we do not support 13 we do not need the logic that says all versions before 15 had identical output.

artemgavrilov: Ah, missed this change!

artemgavrilov (approved):

PR #778, opened 2026-07-09
Decouple structure for local statistics accumulation and hashmap entry

artemgavrilov:
While these two structures have a lot in common, they serve different purposes and have different needs. Decoupling them makes code easier to understand on both ends. 

Also we can split `pgsm_update_entry` function in two, first one `pgsm_update_counter` just to update local counters and the second `pgsm_merge_counters` merges local counter into final hashmap entry counters.

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -133,7 +133,7 @@ static int	num_relations;		/* Number of relation in the query */
 static bool system_init = false;
 static struct rusage rusage_start;
 
-/* Application name and length; set each time when an entry is created locally */
+/* Application name and length; set each time when an stats is created locally */
 static char app_name[APPLICATIONNAME_LEN];
 
 /* Query buffer, store queries' text. */
@@ -209,10 +209,25 @@ static Datum intarray_get_datum(const int32 *arr, int len);
 
 static int64 pgsm_hash_string(const char *str, int len);
 
-static pgsmEntry *pgsm_create_hash_entry(int64 queryid, const PlanInfo *plan_info);
-static void pgsm_add_to_list(pgsmEntry *entry, const char *query_text, int query_len);
-static void pgsm_delete_entry(uint64 queryid);
-static pgsmEntry *pgsm_get_entry_for_query(int64 queryid, const PlanInfo *plan_info, const char *query_text, CmdType cmd_type);
+
+/*
+ * Structure to accumulate query statistics in local memory before putting them into bucket.
+ */
+typedef struct pgsmQueryStats
+{
+	pgsmHashKey key;			/* key used to find/create the shared
+								 * pgsmEntry */
+	int64		pgsm_query_id;	/* pgsm generate normalized query hash */
+	char	   *query;			/* query text, palloc'd in local context */
+	char		datname[NAMEDATALEN];	/* database name */
+	char		username[NAMEDATALEN];	/* user name */
+	Counters	counters;		/* the statistics for this query */
+} pgsmQueryStats;
+
+static pgsmQueryStats *pgsm_create_query_stats(int64 queryid, const PlanInfo *plan_info);
+static void pgsm_add_query_stats(pgsmQueryStats *stats, const char *query_text, int query_len);
+static void pgsm_delete_query_stats(uint64 queryid);
+static pgsmQueryStats *pgsm_get_query_stats(int64 queryid, const PlanInfo *plan_info, const char *query_text, CmdType cmd_type);
 static int64 get_pgsm_query_id_hash(const char *norm_query, int len);
 
 static void pgsm_cleanup_callback(void *arg);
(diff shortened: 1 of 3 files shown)


src/pg_stat_monitor.c
```diff
@@ -215,12 +208,13 @@ static int64 pgsm_hash_string(const char *str, int len);
...
-	int64		pgsm_query_id;			/* pgsm generated normalized query hash */
-	char	   *query;					/* query text, palloc'd in local context */
+	pgsmHashKey key;			/* key used to find/create the shared
+								 * pgsmEntry */
+	int64		pgsm_query_id;	/* pgsm generated normalized query hash */
+	char	   *query;			/* query text, palloc'd in local context */
 	char		datname[NAMEDATALEN];	/* database name */
 	char		username[NAMEDATALEN];	/* user name */
-	Counters	counters;				/* the statistics for this query */
+	Counters	counters;		/* the statistics for this query */
```
jeltz, line 217: Why did you change the comment alignment? This does not seem related to the rest of the commit.

src/pg_stat_monitor.c
```diff
@@ -720,9 +735,9 @@ pgsm_ExecutorEnd(QueryDesc *queryDesc)
 		sys_info.utime = time_diff(rusage_end.ru_utime, rusage_start.ru_utime);
 		sys_info.stime = time_diff(rusage_end.ru_stime, rusage_start.ru_stime);
 
-		entry->counters.info.cmd_type = queryDesc->operation;
+		stats->counters.info.cmd_type = queryDesc->operation;
 
-		pgsm_update_entry(entry,	/* entry */
+		pgsm_update_counters(&stats->counters, stats->key.parentid,	/* counters, parentid */
```
jeltz: Put parentid on a separate row. It is very easy to miss like this.

src/pg_stat_monitor.c
```diff
@@ -1230,7 +1245,8 @@ pg_get_client_addr(void)
 }
 
 static void
-pgsm_update_entry(pgsmEntry *entry,
+pgsm_update_counters(Counters *counters,
+				  int64 parentid,
```
jeltz: You should re-indent the arguments here.

src/pg_stat_monitor.c
```diff
@@ -1838,57 +1863,57 @@ pgsm_store(pgsmEntry *entry)
...
-		shared_hash_entry->stats_since = reset_timestamp;
-		shared_hash_entry->minmax_stats_since = reset_timestamp;
+		memset(&entry->counters, 0, sizeof(Counters));
+		entry->counters.info.cmd_type = stats->counters.info.cmd_type;
+		entry->stats_since = reset_timestamp;
+		entry->minmax_stats_since = reset_timestamp;
 	}
 
-	pgsm_update_entry(shared_hash_entry,	/* entry */
+	pgsm_update_counters(&entry->counters, entry->key.parentid,	/* counters, parentid */
```
jeltz: I think parentid belongs on an own line.

jeltz (changes requested):
I have not fully reviewed the second commit but the first commit is straightforward and a really good change.

src/pg_stat_monitor.c
```diff
@@ -1246,81 +1234,24 @@ pg_get_client_addr(void)
...
+					 uint64 rows,
+					 const BufferUsage *bufusage,
+					 const WalUsage *walusage,
+					 const struct JitInstrumentation *jitusage,
+					 int parallel_workers_to_launch,
+					 int parallel_workers_launched)
 {
-	if (kind == PGSM_PLAN || kind == PGSM_STORE)
-	{
-		counters->plancalls.calls += 1;
```
jeltz: Is it really safe to remove this? What is a plan is planned several times before the first eexecute? Not necessarily saying the current behavior is good or that this is a common scenario but isn't this a behavioral change?

Same with min_max.

But I wonder if this is alos not a bug. Why should the number of calls be incremented both on store and on plan?

src/pg_stat_monitor.c
```diff
@@ -1415,21 +1296,6 @@ pgsm_update_counters(Counters *counters,
...
-		counters->blocks.instr_local_blk_read_time = bufusage->local_blk_read_time;
-		counters->blocks.instr_local_blk_write_time = bufusage->local_blk_write_time;
-#else
-		counters->blocks.instr_shared_blk_read_time = bufusage->blk_read_time;
-		counters->blocks.instr_shared_blk_write_time = bufusage->blk_write_time;
-#endif
-
-#if PG_VERSION_NUM >= 150000
-		counters->blocks.instr_temp_blk_read_time = bufusage->temp_blk_read_time;
-		counters->blocks.instr_temp_blk_write_time = bufusage->temp_blk_write_time;
```
jeltz: Not sure if it can be done but I feel it would be nice if we could handle these in a separate commit. Also aren't they super weird in that they do not record sum or min/max but just the latest value?

src/pg_stat_monitor.c
```diff
@@ -1468,24 +1334,171 @@ pgsm_update_counters(Counters *counters,
...
+	/* copy the plan text once */
+	if (!dst->planinfo.plan_text[0])
+	{
+		dst->planinfo.planid = src->planinfo.planid;
+		dst->planinfo.plan_len = src->planinfo.plan_len;
+		strlcpy(dst->planinfo.plan_text, src->planinfo.plan_text, PLAN_TEXT_LEN);
+	}
+
+	/* copy the query metadata once */
+	if (pgsm_extract_comments && comments && comments[0] && !dst->info.comments[0])
```
jeltz: Wouldn't it make more sense to put all of these, (app_name, comments, ...) directly in pgsm_store() as they do not really relate to counters? Maybe that could also be done as a separate commit?

jeltz (changes requested):
I think we could split the second commit into several smaller, at least two smaller. But I could of course be wrong.

src/pg_stat_monitor.c
artemgavrilov: Accidentally, reverted

src/pg_stat_monitor.c

src/pg_stat_monitor.c

src/pg_stat_monitor.c

src/pg_stat_monitor.c
artemgavrilov: Yes, all these calculations were just dead code because in pgsm_store we passed only totals to hash entry. So there is no behavior change, but this is a question if planner hook can be invoked multiple times in the scope of one query?

https://github.com/percona/pg_stat_monitor/blob/4ba2a1499593b3f9a25685d12a5a6eb46c83d734/src/pg_stat_monitor.c#L1872-L1873

src/pg_stat_monitor.c
artemgavrilov: I thought about it, but for some reasons comments stored under `counters.info.comments`

I did the changes, but later we probably should revisit counters structure.

src/pg_stat_monitor.c
artemgavrilov: I tried to split the commit - I failed. These changes too coupled.

> Also aren't they super weird in that they do not record sum or min/max but just the latest value?
Mmm, yes.

src/pg_stat_monitor.c
jeltz: If they were dead code it would be nice to rome that in a separate commit.

artemgavrilov: "dead code" probably not a right term here. This code actually makes sense only when pgsm_update invoked inside pgsm_store. In all other invocations of pgsm_update these calculations has no sense, their result omitted. Again, because it's spaghetti code it's hard to untangle it by portions.

jeltz (approved):

PR #782, opened 2026-07-11
Simplify shared memory size calculations

jeltz:
We calculated the shared memory size in an unnecessarily convoluted way. I am still a bit skeptical we should access GUCs in macros but this is at least a bit easier to read.

PG-0

### DIFF
--- a/src/hash_query.c
+++ b/src/hash_query.c
@@ -21,12 +21,6 @@
 #include "guc.h"
 #include "hash_query.h"
 
-#define MAX_QUERY_BUF			((int64) pgsm_query_shared_buffer * 1024 * 1024)
-#define MAX_BUCKETS_MEM 		((int64) pgsm_max * 1024 * 1024)
-#define MAX_BUCKET_ENTRIES 		(MAX_BUCKETS_MEM / sizeof(pgsmEntry))
-#define PGSM_BUCKET_INFO_SIZE	(sizeof(TimestampTz) * pgsm_max_buckets)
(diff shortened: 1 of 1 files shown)


src/hash_query.c
```diff
@@ -21,11 +21,8 @@
 #include "guc.h"
 #include "hash_query.h"
 
-#define MAX_QUERY_BUF			((int64) pgsm_query_shared_buffer * 1024 * 1024)
-#define MAX_BUCKETS_MEM 		((int64) pgsm_max * 1024 * 1024)
-#define MAX_BUCKET_ENTRIES 		(MAX_BUCKETS_MEM / sizeof(pgsmEntry))
-#define PGSM_BUCKET_INFO_SIZE	(sizeof(TimestampTz) * pgsm_max_buckets)
-#define PGSM_SHARED_STATE_SIZE	(sizeof(pgsmSharedState) + PGSM_BUCKET_INFO_SIZE)
+#define MAX_BUCKET_ENTRIES 		(((int64) pgsm_max * 1024 * 1024) / sizeof(pgsmEntry))
+#define PGSM_SHARED_STATE_SIZE	(sizeof(pgsmSharedState) + sizeof(TimestampTz) * pgsm_max_buckets)
```
artemgavrilov: I think it worth to add comments here. Before macro names highlighted what are the parts of calculations,

jeltz: Not convinced that is the right fix. I think converting these macros into functions would be better.

jeltz: Is this better?

artemgavrilov (approved):

src/hash_query.c
artemgavrilov: Yep

PR #786, opened 2026-07-13
Clean up handling of return system() values in TAP test

jeltz:
Stop printing the return values and use the same variable name for return values from `system()` as we use for ones from `psql()`.

PG-0

### DIFF
--- a/t/032_multiple_extensions.pl
+++ b/t/032_multiple_extensions.pl
@@ -119,12 +119,10 @@ PGSM::append_to_debug_file($stdout);
 # Create example database and run pgbench init
 ($cmdret, $stdout, $stderr) =
   $node->psql('postgres', 'CREATE DATABASE example;', extra_params => ['-a']);
-print "cmdret $cmdret\n";
 is($cmdret, 0, "CREATE DATABASE example");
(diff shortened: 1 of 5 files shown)


t/032_multiple_extensions.pl
```diff
@@ -119,19 +119,15 @@
 # Create example database and run pgbench init
 ($cmdret, $stdout, $stderr) =
   $node->psql('postgres', 'CREATE DATABASE example;', extra_params => ['-a']);
-print "cmdret $cmdret\n";
 is($cmdret, 0, "CREATE DATABASE example");
 PGSM::append_to_debug_file($stdout);
 
 my $port = $node->port;
-print "port $port \n";
```
dutow, line 127: technically this change is unrelated but I guess that's fine, it's kind of similar

dutow (approved):

t/032_multiple_extensions.pl
jeltz: Yeah, I should probably split the commit.

jeltz: Fixed

PR #788, opened 2026-07-13
Fix likely typo in TAP assertion messages

jeltz:
This seems like a typo which accidentally got copied a bunch of times.

PG-0


dutow (commented):
there's a bunch more of this, even in this file (line 67, 73, 85, ...) and also  in others

jeltz:
I thought I had searched for more instances of this typo but apparently I had not. :( Fixed now.

dutow (approved):

PR #789, opened 2026-07-14
Move code from ExecutorCheckPerms into ExecutorStart

jeltz:
The data we get in the `ExecutorCheckPerms` hook is also available in the `QueryDesc` object we get in the other executor hooks so doing the relation name lookup in a separate hook only makes the code more complicated for no gain.

It might make even more sense to move this lookup to the `ExecutorEnd` hook but that risks changing the behavior as we cannot do catalog lookups when the transaction has been rolled back. It is unclear if we need to support that, but that kind of change should be its own commit.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -155,7 +155,6 @@ static ExecutorEnd_hook_type prev_ExecutorEnd = NULL;
 static ProcessUtility_hook_type prev_ProcessUtility = NULL;
 static emit_log_hook_type prev_emit_log_hook = NULL;
 static shmem_startup_hook_type prev_shmem_startup_hook = NULL;
-static ExecutorCheckPerms_hook_type prev_ExecutorCheckPerms_hook = NULL;
 
 #if PG_VERSION_NUM >= 150000
 static void pgsm_shmem_request(void);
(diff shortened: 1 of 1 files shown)


artemgavrilov (approved):

src/pg_stat_monitor.c
```diff
@@ -567,6 +562,56 @@ pgsm_ExecutorStart(QueryDesc *queryDesc, int eflags)
 			queryDesc->totaltime = InstrAlloc(1, INSTRUMENT_ALL, false);
 			MemoryContextSwitchTo(oldcxt);
 		}
+
+		/*
+		 * Save names of relations involved in the qeury.
```
dutow, line 567: typo

dutow (commented):
This introduced a bug (or two, depending on how count it): the utility path doesn't update `relations` and `num_relations`

```sql
-- Observed results for the trailing COPY:
--   base (main)     : {public.logtbl}          (correct — COPY's own table)
--   patched (pr789) : {public.sensitive_audit} (WRONG — from the prior errored query)
CREATE EXTENSION IF NOT EXISTS pg_stat_monitor;
CREATE TABLE sensitive_audit(id int);
INSERT INTO sensitive_audit VALUES (1);
CREATE TABLE logtbl(id int);
SELECT pg_stat_monitor_reset();

-- Tracked query touching sensitive_audit, errors during ExecutorRun:
SELECT id, 1/(id-1) FROM sensitive_audit;   -- division by zero
-- Unrelated utility statement stored via the ProcessUtility path:
COPY logtbl TO '/tmp/pgsm_stale.csv' CSV;

SELECT substring(query for 45) AS query, relations
FROM pg_stat_monitor
WHERE query ILIKE '%COPY logtbl%' OR query ILIKE '%sensitive%'
ORDER BY query;
```

and

```sql
-- Observed results:
--                                      base (main)                 patched (pr789)
--   SELECT ... FROM parent             {public.parent}             {public.parent}   (unchanged)
--   INSERT INTO child (has FK)         {}  (clobbered/empty)       {public.child}    (FIXED)
--   UPDATE child (has FK)              {}  (clobbered/empty)       {public.child}    (FIXED)
--   COPY parent TO ...                 {public.parent}             {}                (REGRESSION)

CREATE EXTENSION IF NOT EXISTS pg_stat_monitor;
CREATE TABLE parent(id int primary key);
CREATE TABLE child(id int primary key, pid int references parent(id));
INSERT INTO parent VALUES (1),(2),(3);
SELECT pg_stat_monitor_reset();

SELECT * FROM parent WHERE id = 1;
INSERT INTO child VALUES (10, 1);
COPY parent TO '/tmp/pgsm_copy_test.csv' CSV;
UPDATE child SET pid = 2 WHERE id = 10;

SELECT substring(query for 42) AS query, relations
FROM pg_stat_monitor
WHERE query NOT ILIKE '%pg_stat_monitor%' AND query NOT ILIKE '%CREATE%'
ORDER BY query;
````

PR #793, opened 2026-07-15
PG-2557 Read utility statement exec info before executing this statement

artemgavrilov:
PG-2557

There are two reasons for that change:

1. As a preparation for memory management rework we should move all allocations under utility statement processing as it may reset transaction context that we plan to use as a parent for local PGSM memory context.
2. Some data may be collected only before utility statement execution as statement itself may change this data. Because of that make snapshot before executing statement and later use it for stats item creation.

### DIFF
--- a/regression/expected/rollback.out
+++ b/regression/expected/rollback.out
@@ -26,7 +26,7 @@ SELECT pg_stat_monitor_reset();
  
 (1 row)
 
--- During manual rollback we will be able to get username and datname of the query that was rolled back.
+-- During manual rollback we will be able to get username of the query that was rolled back.
 BEGIN;
 INSERT INTO t VALUES (2);
 ROLLBACK;
@@ -45,7 +45,7 @@ SELECT pg_stat_monitor_reset();
  
 (1 row)
 
--- If transaction is rolled back due an error, we won't be able to get username and datname of the query that was rolled back.
+-- If transaction is rolled back due an error, we won't be able to get username of the query that was rolled back.
 BEGIN;
 INSERT INTO t VALUES (1);
 ERROR:  duplicate key value violates unique constraint "t_pkey"
@@ -56,7 +56,7 @@ SELECT query, username, datname FROM pg_stat_monitor ORDER BY query COLLATE "C";
 --------------------------------+----------+--------------------
  BEGIN                          | u        | contrib_regression
  INSERT INTO t VALUES (1);      | u        | contrib_regression
- ROLLBACK                       |          | 
+ ROLLBACK                       |          | contrib_regression
  SELECT pg_stat_monitor_reset() | u        | contrib_regression
 (4 rows)
 
--- a/regression/expected/security_definer.out
+++ b/regression/expected/security_definer.out
@@ -1,0 +1,139 @@
+CREATE USER su WITH SUPERUSER;
+SET ROLE su;
+CREATE EXTENSION pg_stat_monitor;
+SET pg_stat_monitor.pgsm_track = 'all';
+CREATE USER owner_role;
+CREATE USER caller_role;
+GRANT ALL ON SCHEMA public TO owner_role;
+SET ROLE owner_role;
+CREATE TABLE owner_table (a int);
+INSERT INTO owner_table VALUES (1);
+CREATE FUNCTION owner_secdef_func() RETURNS bigint AS $$
+BEGIN
+    RETURN (SELECT count(*) FROM owner_table);
+END
+$$ LANGUAGE plpgsql SECURITY DEFINER;
+CREATE FUNCTION owner_secdef_ddl_func() RETURNS void AS $$
+BEGIN
+    EXECUTE 'CREATE TABLE owner_ddl_table (a int)';
+    EXECUTE 'DROP TABLE owner_ddl_table';
+END
+$$ LANGUAGE plpgsql SECURITY DEFINER;
+CREATE FUNCTION owner_secdef_appname_func() RETURNS void AS $$
+BEGIN
+    PERFORM set_config('application_name', 'secdef_app', false);
+    PERFORM 2 + 2;
+END
+$$ LANGUAGE plpgsql SECURITY DEFINER;
+GRANT EXECUTE ON FUNCTION owner_secdef_func() TO caller_role;
+GRANT EXECUTE ON FUNCTION owner_secdef_ddl_func() TO caller_role;
+GRANT EXECUTE ON FUNCTION owner_secdef_appname_func() TO caller_role;
+SET ROLE su;
+SELECT pg_stat_monitor_reset();
+ pg_stat_monitor_reset 
+-----------------------
+ 
+(1 row)
+
+SET ROLE caller_role;
+SELECT owner_secdef_func();
+ owner_secdef_func 
+-------------------
+                 1
+(1 row)
+
+SET ROLE su;
+-- The outer call must be attributed to the caller, while the query executed
+-- inside the SECURITY DEFINER function body must be attributed to the
+-- function owner, since that is whose privileges it runs under.
+SELECT username, top_query, query FROM pg_stat_monitor
+  WHERE query NOT LIKE '%pg_stat_monitor%'
+  ORDER BY username, query COLLATE "C";
+  username   |          top_query          |               query                
+-------------+-----------------------------+------------------------------------
+ caller_role |                             | SELECT owner_secdef_func()
+ caller_role |                             | SET ROLE su
+ owner_role  | SELECT owner_secdef_func(); | (SELECT count(*) FROM owner_table)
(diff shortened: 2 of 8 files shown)


src/pg_stat_monitor.c
```diff
@@ -1043,6 +1046,8 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 		}
 		PG_END_TRY();
 
+		stats = pgsm_create_query_stats(queryId, NULL);
```
artemgavrilov: Main idea of this PR is to move local stat item creation back under utility processing because some utility statements (VACUUM for example) create and destroy transactions internally.  So if we will create item before it will be wiped out by statement execution.

src/pg_stat_monitor.c
```diff
@@ -1780,8 +1784,8 @@ pgsm_store(const pgsmQueryStats *stats)
 	if (pgsm_extract_comments && comments[0] && !entry->counters.info.comments[0])
 		strlcpy(entry->counters.info.comments, comments, COMMENTS_LEN);
 
-	if (pgsm_track_application_names && app_name[0] != '\0' && !entry->counters.info.application_name[0])
-		strlcpy(entry->counters.info.application_name, app_name, APPLICATIONNAME_LEN);
+	if (pgsm_track_application_names && cached_appname[0] != '\0' && !entry->counters.info.application_name[0])
```
artemgavrilov: BTW probably it makes sense to deprecate `pgsm_track_application_names` parameter.

src/pg_stat_monitor.c
```diff
@@ -1168,21 +1172,53 @@ pgsm_hash_string(const char *str, int len)
...
+pgsm_update_cached_info()
 {
-	int			len;
+	/*
+	 * Get the user ID. Let's use this instead of GetUserID as this won't
+	 * throw an assertion in case of an error.
+	 */
+	int			sec_ctx;
+
+	GetUserIdAndSecContext(&cached_userid, &sec_ctx);
```
dutow: do we have to do user name lookup possibly multiple times for a query?

dutow (commented):
Isn't `pgsm_store_error` handling missing?

src/pg_stat_monitor.c
```diff
@@ -126,8 +124,12 @@ static int	num_relations;		/* Number of relation in the query */
...
 static struct rusage rusage_start;
 
-/* Application name and length; set each time when an stats is created locally */
-static char app_name[APPLICATIONNAME_LEN];
+/* Cached per-backend information */
+static char cached_appname[NAMEDATALEN];
+static char cached_datname[NAMEDATALEN];
+static char cached_username[NAMEDATALEN];
+static Oid	cached_userid = InvalidOid;
+static uint32 cached_client_ip = PGSM_INVALID_IP;
```
jeltz: Not sure I like this idea. What about for example `SECURITY DEFINER` functions which contain queries. Will this not break logging them as the right user?

src/pg_stat_monitor.c
jeltz: Yeah, we should deprecate it.

jeltz (changes requested):
Forgot to update status. As I wrote in my comment I am not a fan of this direction. I think your alternate proposal you mentioned on Slack make more sense.

src/pg_stat_monitor.c
artemgavrilov: Will do in separate PR

src/pg_stat_monitor.c
```diff
@@ -1013,10 +1025,18 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+		pgsmSnapshot snapshot;
 
 		getrusage(RUSAGE_SELF, &rusage_start);
 
+		/*
+		 * Create a snapshot before utility statement execution because
+		 * statement may change the data itself and for statistics we need to
+		 * know this data at the time of the statement execution.
+		 */
+		pgsm_create_snapshot(&snapshot);
```
artemgavrilov: Mmm, probably not the best name.

artemgavrilov:
@jeltz @dutow after slack discussion I reworked this PR to completely different solution.

src/pg_stat_monitor.c
jeltz: Was going to point out the same thing!

src/pg_stat_monitor.c
```diff
@@ -1168,21 +1190,58 @@ pgsm_hash_string(const char *str, int len)
...
+	 * throw an assertion in case of an error.
+	 */
+	GetUserIdAndSecContext(&snapshot->userid, &sec_ctx);
 
 	if (application_name && *application_name)
-		len = strlcpy(name, application_name, buff_size);
+		strlcpy(snapshot->appname, application_name, NAMEDATALEN);
 	else
-		len = strlcpy(name, "unknown", buff_size);
+		strlcpy(snapshot->appname, "unknown", NAMEDATALEN);
```
jeltz: We should probably fix this in a separate PR. This unknown makes no sense. :D

regression/expected/rollback.out
```diff
@@ -56,7 +56,7 @@ SELECT query, username, datname FROM pg_stat_monitor ORDER BY query COLLATE "C";
 --------------------------------+----------+--------------------
  BEGIN                          | u        | contrib_regression
  INSERT INTO t VALUES (1);      | u        | contrib_regression
- ROLLBACK                       |          | 
+ ROLLBACK                       |          | contrib_regression
```
jeltz, line 59: We probably want some more tests than this given how your previous version of the pacth likely broke some stuff which was not caught by our test suite.

jeltz (changes requested):
This looks good but I think we also want to move the collection of this data to the ExecutorStart hook as a future improvement in a separate PR. Sure, then we would need to separate the snapshot from the stats object but that is not necessarily a bad thing. But this PR is big improvement and does not need to fix those things too.

jeltz:
Also this PR should likely have Jira issue and a changelog entry now that it actually fixes some issues.

src/pg_stat_monitor.c
artemgavrilov: In new PR version it will be once per stats item creation.

artemgavrilov:
> Also this PR should likely have Jira issue and a changelog entry now that it actually fixes some issues.

Not sure, it doesn't change anything in current behavior, but prepares for memory context rework. 

> move the collection of this data to the ExecutorStart hook as a future improvement in a separate PR

Yes, was planning to do that as part of memory rework

src/pg_stat_monitor.c
artemgavrilov: That may be `exec_info` instead of snapshot. And func name will be `pgsm_fill_exec_info()`. What do you think?

jeltz:
> Not sure, it doesn't change anything in current behavior, but prepares for memory context rework.

Well, there is a diff in our tests so I would say it changes something.

src/pg_stat_monitor.c
jeltz: Works for me.

artemgavrilov:
> > Not sure, it doesn't change anything in current behavior, but prepares for memory context rework.
> 
> Well, there is a diff in our tests so I would say it changes something.

Created jira ticket PG-2557 and add changelog record

regression/expected/rollback.out
artemgavrilov: Added test

regression/expected/security_definer.out
```diff
@@ -0,0 +1,139 @@
...
+-----------------------
+ 
+(1 row)
+
+SET ROLE su;
+-- DDL executed inside a SECURITY DEFINER function must still be
+-- attributed to the function owner, not the caller.
+SELECT username, top_query, query FROM pg_stat_monitor
+  WHERE query NOT LIKE '%pg_stat_monitor%'
+  ORDER BY username, query COLLATE "C";
```
jeltz, line 78: I wonder if not ordering by `stats_since` would make the tests easier to read.

jeltz (approved):
Had a small comment that you optionally can address.

regression/expected/security_definer.out
artemgavrilov: I've tried, nested queries in that case logged in backward order, and that may be confusing at first too:

```
  username   |          top_query          |               query                
-------------+-----------------------------+------------------------------------
 su          |                             | SET ROLE caller_role
 owner_role  | SELECT owner_secdef_func(); | (SELECT count(*) FROM owner_table) <--- inner
 caller_role |                             | SELECT owner_secdef_func() <--- outer
 caller_role |                             | SET ROLE su
(4 rows)

```

artemgavrilov: Do you still prefer it that way?

jeltz: nope!

PR #797, opened 2026-07-21
PG-2229 Rework memory management

artemgavrilov:
PG-2229

Use TopTransactionContext as a parent for pgsm context. Create it lazy for each statement. If there were aborted subtransactions, cleanup local stats list for such entries, otherwise statistics may be compromised. For statement error scenario and utility statements we don't need cross-hook objects, so allocate stats object there on stack.

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -225,9 +225,8 @@ typedef struct pgsmQueryExecInfo
 
 static MemoryContext pgsm_memory_context(void);
 static void pgsm_fill_query_exec_info(pgsmQueryExecInfo *info);
-static pgsmQueryStats *pgsm_create_query_stats(int64 queryid, int64 planid, int64 pgsm_query_id, const char *query_text, CmdType cmd_type);
+static pgsmQueryStats *pgsm_add_query_stats(int64 queryid, int64 planid, int64 pgsm_query_id, const char *query_text, int query_len, CmdType cmd_type);
 static void pgsm_fill_query_stats(pgsmQueryStats *stats, const pgsmQueryExecInfo *info, int64 queryid, int64 planid, int64 pgsm_query_id, const char *query_text, CmdType cmd_type);
-static void pgsm_add_query_stats(pgsmQueryStats *stats, const char *query_text, int query_len);
 static void pgsm_delete_query_stats(uint64 queryid);
 static pgsmQueryStats *pgsm_get_query_stats(int64 queryid, int64 planid, const char *query_text, CmdType cmd_type);
 static int64 get_pgsm_query_id_hash(const char *norm_query, int len);
@@ -415,7 +414,6 @@ pgsm_shmem_request(void)
 static void
 pgsm_post_parse_analyze_internal(ParseState *pstate, Query *query, JumbleState *jstate)
 {
-	pgsmQueryStats *stats;
 	const char *query_text;
 	int			query_len;
 	const char *hash_text;
@@ -495,10 +493,9 @@ pgsm_post_parse_analyze_internal(ParseState *pstate, Query *query, JumbleState *
 	 * bucket value. The correct bucket value will be needed then to search
 	 * the hash table, or create the appropriate entry.
 	 */
-	stats = pgsm_create_query_stats(query->queryId, 0,
-									get_pgsm_query_id_hash(hash_text, hash_text_len),
-									store_text, query->commandType);
-	pgsm_add_query_stats(stats, store_text, store_text_len);
+	pgsm_add_query_stats(query->queryId, 0,
+						 get_pgsm_query_id_hash(hash_text, hash_text_len),
+						 store_text, store_text_len, query->commandType);
 
 	Assert(list_length(lentries) <= max_nesting_level);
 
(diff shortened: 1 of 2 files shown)


src/pg_stat_monitor.c
```diff
@@ -1100,12 +1086,13 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 
 		query_text = CleanQuerytext(queryString, &location, &query_len);
 
-		stats->query = pnstrdup(query_text, query_len);
-		stats->pgsm_query_id = get_pgsm_query_id_hash(query_text, query_len);
-		stats->counters.info.cmd_type = cmd_type;
+		pgsm_fill_query_stats(&stats, &info, queryId, 0);
+		stats.query = pnstrdup(query_text, query_len);
```
jeltz: Can't we use unconstify() here to?

src/pg_stat_monitor.c
```diff
@@ -1485,58 +1471,101 @@ pgsm_merge_counters(Counters *dst, const Counters *src)
...
+	Assert(IsTransactionState());
 
-	pgsm_store(stats);
+	if (PgsmMemoryContext == NULL)
+	{
+		/*
+		 * We expect top transaction context to be available in any possible
+		 * scenario. CurrentMemoryContext here is just a failsafe mechanism,
+		 * it should never happen.
+		 */
```
jeltz, line 1486: Should we have an assert here so it crashes in assertion enabled builds?

src/pg_stat_monitor.c
```diff
@@ -1485,58 +1471,101 @@ pgsm_merge_counters(Counters *dst, const Counters *src)
...
-	stats = palloc0_object(pgsmQueryStats);
+	stats->subxid = IsTransactionState() ? GetCurrentSubTransactionId()
+		: InvalidSubTransactionId;
 
 	if (pgsm_track_application_names)
 	{
 		stats->key.appid = pgsm_hash_string(info->appname, strlen(info->appname));
 	}
 	stats->key.ip = client_ip;
-	stats->key.planid = plan_info ? plan_info->planid : 0;
```
jeltz, line 1523: Isn't this change unrelated to the rest of the commit? It is not necessarily bad but it seems quite unrelated.

src/pg_stat_monitor.c
```diff
@@ -1655,11 +1679,38 @@ pgsm_get_query_stats(int64 queryid, const PlanInfo *plan_info, const char *query
...
+ * On subtransaction abort, discard any in-flight local stats that belong to the
+ * aborting subtransaction.
+ */
+static void
+pgsm_subxact_callback(SubXactEvent event, SubTransactionId mySubid,
+					  SubTransactionId parentSubid, void *arg)
+{
+	ListCell   *lc;
+
+	if (event != SUBXACT_EVENT_ABORT_SUB || lentries == NIL)
```
jeltz: I feel `lentries == NIL` is quite an unnecessary optimization as I am sure the compiler can handle it just fine.

src/pg_stat_monitor.c
```diff
@@ -1086,10 +1080,12 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 
 		query_text = CleanQuerytext(queryString, &location, &query_len);
 
-		pgsm_fill_query_stats(&stats, &info, queryId, 0);
-		stats.query = pnstrdup(query_text, query_len);
-		stats.pgsm_query_id = get_pgsm_query_id_hash(query_text, query_len);
-		stats.counters.info.cmd_type = cmd_type;
+		/* Make it null terminated */
+		store_text = pnstrdup(query_text, query_len);
```
jeltz, line 1068: Ah, so that is why it was not fixed in the previous commit. Hmm, maybe this comment should then be added in the previous commit.

src/pg_stat_monitor.c
```diff
@@ -1571,6 +1561,10 @@ pgsm_fill_query_stats(pgsmQueryStats *stats, const pgsmQueryExecInfo *info, int6
 	stats->key.parentid = 0;
 	stats->key.userid = info->userid;
 
+	stats->pgsm_query_id = pgsm_query_id;
+	stats->counters.info.cmd_type = cmd_type;
+	stats->query = unconstify(char *, query_text);
```
jeltz, line 1550: What would happen if you made the field a `const char *`? Would you get any error?

src/pg_stat_monitor.c
```diff
@@ -1518,23 +1515,26 @@ pgsm_memory_context(void)
...
 
-	/* Create the stats in the pgsm memory context */
+	/* Create the stats and own the query text in the pgsm memory context */
 	oldctx = MemoryContextSwitchTo(pgsm_memory_context());
 	stats = palloc0_object(pgsmQueryStats);
-	MemoryContextSwitchTo(oldctx);
+	store_text = pnstrdup(query_text, query_len);
 
-	pgsm_fill_query_stats(stats, &info, queryid, planid, pgsm_query_id, query_text, cmd_type);
+	pgsm_fill_query_stats(stats, &info, queryid, planid, pgsm_query_id, store_text, cmd_type);
```
jeltz, line 1519: Wouldn't it be cleaner to call this outside the memory context scope? To make it clear no allocations happen inside it.

jeltz (changes requested):
Looks good but I had some comments.

src/pg_stat_monitor.c
artemgavrilov: We have it, at the very top of this function. But assert will trigger only in test builds and here I'm trying to prevent any possible prod crashes.

src/pg_stat_monitor.c
artemgavrilov: Done

src/pg_stat_monitor.c
artemgavrilov: You're right. Removed.

src/pg_stat_monitor.c
artemgavrilov: Fair. Extracted into separate commit.

src/pg_stat_monitor.c
artemgavrilov: It will then require `unconstify` in many other places, for example for `free(stats->query)` calls

src/pg_stat_monitor.c
artemgavrilov: We need memory context right above stats and query_text allocation, and right after for `lappend` call. Technically we can first put stats in the list, and then fill it. But as for me it will break natural order. Other option is to switch context twice, but not sure that it worth it

src/pg_stat_monitor.c
artemgavrilov: As you mentioned later in this PR, this was done to get nul terminated string

jeltz (approved):

PR #801, opened 2026-07-24
Clean up get_next_wbucket()

jeltz:
The function had an overly complex flow and in general made things more complex than necessary so we simplify the loop where we wait for the atomic, change to using `pg_atomic_write_u64()`, only assign the bucket start once, avoid changing the `timeval` and rename some variables.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -1994,7 +1994,7 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
 	hash_seq_init(&hstat, get_pgsmHash());
 
 	now = GetCurrentTimestamp();
-	current_bucket = pg_atomic_read_u64(&pgsm->current_wbucket);
+	current_bucket = pg_atomic_read_u64(&pgsm->current_bucket_id);
 
 	while ((entry = hash_seq_search(&hstat)) != NULL)
 	{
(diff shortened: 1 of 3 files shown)


artemgavrilov (approved):

src/pg_stat_monitor.c
```diff
@@ -2471,69 +2445,61 @@ static uint64
...
+		if (pg_atomic_compare_exchange_u64(&pgsm->current_bucket_start, &current_bucket_start, (uint64) tv.tv_sec))
+			break;
+	}
 
-		new_bucket_id = (tv.tv_sec / pgsm_bucket_time) % pgsm_max_buckets;
+	new_bucket_id = (tv.tv_sec / pgsm_bucket_time) % pgsm_max_buckets;
 
-		/* Update bucket id and retrieve the previous one. */
-		pg_atomic_exchange_u64(&pgsm->current_wbucket, new_bucket_id);
+	pg_atomic_write_u64(&pgsm->current_bucket_id, new_bucket_id);
```
dAdAbird, line 2483: I understand it's not related to this PR (and I see a comment below), but all these atomics don't make any sense to me here. There are two atomic vars which kinda depend on each other but two atomic calls is not an atomic operation itself. For example:
1. We read `current_bucket_start` in L2472 and then do an atomic read return the `current_bucket_id` in L2475 based on some logic in L2474. But another thread can update and return `current_bucket_id` in between L2472-75, making two processes write to the same bucket  (? - or whatever they do with the get_next_wbucket() result).
2. Update of `current_bucket_start` in "loop + L2472 and L2477" seems legit. But then againg an atomic update of `current_bucket_id` in L2483, which logically depends on the `current_bucket_start` value, but anything can happen in between those atomic ops.

This should be guarded by RW_LOCK. Atomics are useless here. Or the whole data layout and algorithms should be reconsidered if atomics are necessary here (and I'm not sure if there is a solution).

Again, it's not something that was introduced by this PR, and I probably won't say anything new, though it looks scary to me. Maybe I don't see the bigger picture and am lacking the understanding of the later usage of current_bucket_start/current_bucket_id, but those atomics look wrong to me.

jeltz: Yup, I should have added a comment explaining this. This PR started out as part of an attempt to fix these issues but I paused that and just submitted the cleanup part as this PR.

PR #802, opened 2026-07-24
PG-2602 Deprecate pgsm_track_application_names parameter

artemgavrilov:
PG-2602

This parameter was introduced to address performance issues caused by application name tracking. Code parts that were causing these issues were removed in b0ebcc8, so this parameter has no sense anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Set unit for `pg_stat_monitor.pgsm_query_max_len` to bytes
 - Do not leave the `pgsm_create_view()` function after running `CREATE EXTENSION`
 - Specifying `USE_PGXS` is no longer necessary when building with make
+- Deprecate the `pgsm_track_application_names` parameter, application name now tracked always ([PG-2602](https://perconadev.atlassian.net/browse/PG-2602))
 
 ### Removed
 
--- a/regression/expected/application_name.out
+++ b/regression/expected/application_name.out
@@ -24,32 +24,9 @@ SELECT pg_stat_monitor_reset();
  
 (1 row)
 
-SELECT 1 AS num;
- num 
------
-   1
-(1 row)
-
+-- The parameter is deprecated and has no effect, but setting it must still
+-- emit a deprecation warning.
 SET pg_stat_monitor.pgsm_track_application_names = off;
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 - Do not leave the `pgsm_create_view()` function after running `CREATE EXTENSION`
 - Read utility statement exec info before executing this statement ([PG-2557](https://perconadev.atlassian.net/browse/PG-2557))
+- Deprecate the `pgsm_track_application_names` parameter ([PG-2602](https://perconadev.atlassian.net/browse/PG-2602))
```
jeltz: Maybe we should clarify that they are always tracked now.

src/guc.c
```diff
@@ -181,13 +184,13 @@ init_guc(void)
 		);
 
 	DefineCustomBoolVariable("pg_stat_monitor.pgsm_track_application_names",	/* name */
-							 "Enable/Disable application names tracking.",	/* short_desc */
+							 "Deprecated, has no effect. Application names are always tracked.",	/* short_desc */
 							 NULL,	/* long_desc */
 							 &pgsm_track_application_names, /* value address */
```
jeltz, line 189: Make this variable static and remove it from the header file so it cannot be used.

regression/expected/application_name.out
```diff
@@ -24,32 +24,4 @@ SELECT pg_stat_monitor_reset();
  
 (1 row)
 
-SELECT 1 AS num;
- num 
------
-   1
-(1 row)
-
-SET pg_stat_monitor.pgsm_track_application_names = off;
```
jeltz, line 33: Maybe keep this to check that we get the warning?

jeltz (changes requested):

CHANGELOG.md
artemgavrilov: Added

src/guc.c
artemgavrilov: Done

regression/expected/application_name.out
artemgavrilov: Done

jeltz (approved):

PR #805, opened 2026-07-28
Fix stats for queries executed via prepared statements

artemgavrilov:
If prepared statement was created in one transaction and executed in another it may produce entry with incorrect data if query itself changes application name or user. So to fix this we just need to be sure that stas item exists in executor start hook, were no mutations we done yet.

PG-0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,3 +32,4 @@
 - Potential buffer overread in comment extraction ([PG-2540](https://perconadev.atlassian.net/browse/PG-2540))
 - Do not access `PlannedStmt` after we call `standard_ProcessUtility()` ([PG-2486](https://perconadev.atlassian.net/browse/PG-2486))
 - Various improvements to the stability of our test suite
+- Make sure that for prepared statements utility statement exec info read at the executor start hook, where data is not yet modified by query itself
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -15,3 +15,4 @@
 ### Fixed
 
 - Fix memory leak by scoping local memory context to the current transaction ([PG-2229](https://perconadev.atlassian.net/browse/PG-2229))
+- Fix application and user name tracking for prepared statements
```
jeltz: I wonder if this is not a change. At least that is what we claimed it was for utility commands. We should pick the same classification for both at least.

```
- Read utility statement exec info before executing this statement ([PG-2557](https://perconadev.atlassian.net/browse/PG-2557))
```

jeltz (changes requested):

CHANGELOG.md
artemgavrilov: In the PG-2557 we re-implemented existing behavior. Before it was achieved by side-effects and we changed it to explicit solution. But here we have pure bug. Anyway I changed Changelog record wording.

jeltz (approved):

PR #816, opened 2026-08-07
Avoid heavy operations under spinlock

artemgavrilov:
dsa_allocate() takes the DSA area's LWLock internally. Taking LWLocks under spinlock is fordiden by PostgreSQL and can cause stuck-spinlock panic under contention. So copy parent query to DSA before taking spinlock and in critical section just assign this pointer if needed.

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -1795,6 +1795,9 @@ pgsm_store(const pgsmQueryStats *stats)
 	pgsmHashKey key = stats->key;
 	char	   *query = stats->query;
 	char		comments[COMMENTS_LEN];
+	const char *parent_query = NULL;
+	dsa_pointer parent_query_pointer = InvalidDsaPointer;
+	dsa_area   *query_dsa_area = NULL;
 
 	/* Safety check... */
 	if (!IsSystemInitialized())
@@ -1813,6 +1816,7 @@ pgsm_store(const pgsmQueryStats *stats)
 	if (pgsm_track == PGSM_TRACK_ALL && nesting_level > 0 && nesting_level < max_nesting_level)
 	{
 		key.parentid = nested_queryids[nesting_level - 1];
+		parent_query = nested_query_txts[nesting_level - 1];
 	}
 	else
 	{
@@ -1829,7 +1833,6 @@ pgsm_store(const pgsmQueryStats *stats)
 	if (!entry)
 	{
 		dsa_pointer dsa_query_pointer;
-		dsa_area   *query_dsa_area;
 		char	   *query_buff;
 		int			query_len = strlen(query);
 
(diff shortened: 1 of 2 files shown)


src/pg_stat_monitor.c
```diff
@@ -1920,41 +1949,23 @@ pgsm_store(const pgsmQueryStats *stats)
...
+	/* Store the parent query text unless it's already set */
+	if (DsaPointerIsValid(parent_query_pointer) &&
+		!DsaPointerIsValid(entry->counters.info.parent_query))
 	{
-		Assert(!DsaPointerIsValid(entry->counters.info.parent_query));
+		entry->counters.info.parent_query = parent_query_pointer;
+		parent_query_pointer = InvalidDsaPointer;
 	}
 
+	/* Parent id and query text must be consistent */
```
jeltz: I think the comment can be skipped.

src/pg_stat_monitor.c
```diff
@@ -1920,41 +1949,23 @@ pgsm_store(const pgsmQueryStats *stats)
...
-					memcpy(qry_buff, nested_query_txts[nesting_level - 1], parent_query_len);
-					qry_buff[parent_query_len] = '\0';
-					/* store the dsa pointer for parent query text */
-					entry->counters.info.parent_query = qry;
-				}
-			}
-		}
-	}
-	else
+	/* Store the parent query text unless it's already set */
```
jeltz: This comment is quite useless as-is but can be imrpoved by making the comment explain how we make sure we do not leak memory. The comment could also be at the block where we allocate the DSA.

src/pg_stat_monitor.c
```diff
@@ -1905,6 +1908,32 @@ pgsm_store(const pgsmQueryStats *stats)
...
+		{
+			char	   *parent_query_buff = dsa_get_address(query_dsa_area,
+															parent_query_pointer);
+
+			memcpy(parent_query_buff, parent_query, parent_query_len);
+			parent_query_buff[parent_query_len] = '\0';
+		}
+	}
+
+	/* Keep spinlock section short and fast */
```
jeltz: Quite useless comment. :D No, let's keep it slow and long!

if we feel there needs to be a comment it should be at the block allocating the DSA.

jeltz (changes requested):
Looks good but in general I would want to see comments at a bit higher level rather than retelling what the line below does. One longer comment epxlaining the speculative allocation would be much nicer than the short ones in this patch.

artemgavrilov:
> Looks good but in general I would want to see comments at a bit higher level rather than retelling what the line below does. One longer comment epxlaining the speculative allocation would be much nicer than the short ones in this patch.

Changed to one long comment

jeltz (approved):

PR #820, opened 2026-08-07
Fix plans counter

artemgavrilov:
plans counter was incremented unconditionally, even when there was no plan at all (utility statements). Fix it so it will reflect how many times planning events happened. Also this fixes associated timers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,3 +39,4 @@
 - Make sure that for prepared statements utility statement exec info read at the executor start hook, where data is not yet modified by query itself
 - Do not acquire LWLock under spinlock
 - Race condition where we could leak memory for the parent query
+- `plans` now counts only actual planner invocations instead of every execution: utility statements and executions that reuse a cached plan no longer bump the counter
--- a/regression/expected/planning.out
+++ b/regression/expected/planning.out
@@ -63,8 +63,8 @@ SELECT plans, calls, rows, query FROM pg_stat_monitor
     WHERE query NOT LIKE 'PREPARE%' ORDER BY query COLLATE "C";
  plans | calls | rows |                      query                      
 -------+-------+------+-------------------------------------------------
-     1 |     1 |    0 | ALTER TABLE stats_plan_test ADD COLUMN x int
-     1 |     1 |    0 | CREATE TABLE stats_plan_test ()
+     0 |     1 |    0 | ALTER TABLE stats_plan_test ADD COLUMN x int
+     0 |     1 |    0 | CREATE TABLE stats_plan_test ()
      3 |     3 |    3 | SELECT $1
      1 |     1 |    1 | SELECT pg_stat_monitor_reset() IS NOT NULL AS t
 (4 rows)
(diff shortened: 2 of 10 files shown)


regression/expected/planning.out
```diff
@@ -78,9 +78,58 @@ SELECT plans >= 2 AND plans <= calls AS plans_ok, calls, rows, query FROM pg_sta
...
+
+EXECUTE prep2;
+ ?column? 
+----------
+       42
+(1 row)
+
+RESET plan_cache_mode;
+-- the generic plan is built once and then reused, so a replan forced by a
+-- cache invalidation still leaves fewer plans than calls
```
jeltz: Is this even true? Also how do you imagine a cache invalidation would force a replan? This looks like LLMed nonsense.

regression/expected/planning.out
```diff
@@ -78,9 +78,58 @@ SELECT plans >= 2 AND plans <= calls AS plans_ok, calls, rows, query FROM pg_sta
...
+ ?column? 
+----------
+       42
+(1 row)
+
+EXECUTE prep2;
+ ?column? 
+----------
+       42
+(1 row)
```
jeltz: Why 4 executions?

regression/expected/planning.out
```diff
@@ -78,9 +78,58 @@ SELECT plans >= 2 AND plans <= calls AS plans_ok, calls, rows, query FROM pg_sta
...
+-- the generic plan is built once and then reused, so a replan forced by a
+-- cache invalidation still leaves fewer plans than calls
+SELECT plans > 0 AS planned, plans < calls AS fewer_plans_than_calls,
+       min_plan_time > 0 AS min_plan_time_positive, calls
+    FROM pg_stat_monitor WHERE query LIKE 'PREPARE prep2%'
+    ORDER BY query COLLATE "C";
+ planned | fewer_plans_than_calls | min_plan_time_positive | calls 
+---------+------------------------+------------------------+-------
+ t       | t                      | t                      |     4
+(1 row)
```
jeltz, line 110: Why do you check for that minimum plan time is positive?

regression/expected/planning.out
```diff
@@ -78,9 +78,58 @@ SELECT plans >= 2 AND plans <= calls AS plans_ok, calls, rows, query FROM pg_sta
...
 
+--
+-- Executions reusing a cached plan are not replanned, so they must neither
+-- be counted as plans nor contribute a zero-time sample to the plan timings.
+--
+SELECT pg_stat_monitor_reset() IS NOT NULL AS t;
+ t 
+---
+ t
+(1 row)
```
jeltz, line 88: Not sure why we need to reset stats.

src/pg_stat_monitor.c
```diff
@@ -1455,26 +1460,35 @@ pgsm_merge_counters(Counters *dst, const Counters *src)
...
-		dst->plantime.min_time = src->plantime.total_time;
-		dst->plantime.max_time = src->plantime.total_time;
-		dst->plantime.mean_time = src->plantime.total_time;
-	}
-	else
+	/*
+	 * src may carry more than one planning event (a statement can be planned
+	 * several times before it is executed, e.g. by EXPLAIN); all of them are
+	 * counted, but their accumulated time is still folded as one sample.
+	 */
```
jeltz: Quite useless comment.

src/pg_stat_monitor.c
```diff
@@ -1455,26 +1460,35 @@ pgsm_merge_counters(Counters *dst, const Counters *src)
...
+	if (src->plancalls.calls > 0)
 	{
-		double		old_mean = dst->plantime.mean_time;
+		bool		first_sample = (dst->plancalls.calls == 0);
```
jeltz: Why not just move incrementing `.calls` until later and check if calls is zero directly in the `if`?

CHANGELOG.md
```diff
@@ -39,3 +39,4 @@
 - Make sure that for prepared statements utility statement exec info read at the executor start hook, where data is not yet modified by query itself
 - Do not acquire LWLock under spinlock
 - Race condition where we could leak memory for the parent query
+- The `plans` counter counts how many times a statement was actually planned.
```
jeltz: I think this should be clarified a bit. Also it should not end in full stop.

jeltz (changes requested):

src/pg_stat_monitor.c

src/pg_stat_monitor.c
artemgavrilov: Removed

regression/expected/planning.out
artemgavrilov: To not have collision with previous test case. Simple select and prepared are pretty much the same query and fills the same entry in PGSM.

regression/expected/planning.out
artemgavrilov: Before fix time counter may be decremented down to zero. I reworked this test to check only plan counter, I think that's enough.

regression/expected/planning.out
artemgavrilov: This is indeed LLM slope. I haven't done proper review of that part. Reworked.

regression/expected/planning.out
artemgavrilov: No reason. Changed to 2.

CHANGELOG.md
artemgavrilov: Updated

jeltz (approved):

PR #827, opened 2026-08-10
Only store the upper bound of the histogram buckets

jeltz:
The lower bound is not necessary to store so do not store it and instead only store the upper bounds saving a few bytes and a few cycles.

PG-0

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -114,13 +114,11 @@ static int	plan_nested_level = 0;
 /* Histogram bucket variables */
 static double hist_bucket_min;
 static double hist_bucket_max;
-static struct
-{
-	double		start;
-	double		end;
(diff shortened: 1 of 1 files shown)


src/pg_stat_monitor.c
```diff
@@ -3193,11 +3188,9 @@ histogram_bucket_boundary(int index)
 	 * Can't do exp(0) as that returns 1. So handling the case of first entry
 	 * specifically
 	 */
```
artemgavrilov, line 3193: It seems that comment no longer valid

artemgavrilov (approved):

src/pg_stat_monitor.c
jeltz: It is still as relevant but the comment should be moved.

jeltz: Oh, and the algorthm is stupid.

PR #830, opened 2026-08-18
PG-2623 Utility statement normalization

artemgavrilov:
PG-2623

### Description

To have normalized query text for utility statements we have to pre-save it in post parse hook. (pretty much the same as for ordinary queries). Because some utility statements may cleanup our memory context we have to copy query stats entry to stack.

### DIFF
--- a/src/pg_stat_monitor.c
+++ b/src/pg_stat_monitor.c
@@ -478,10 +478,17 @@ pgsm_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 	 */
 	if (query->utilityStmt)
 	{
-		if (pgsm_track_utility && IsA(query->utilityStmt, ExecuteStmt))
-			query->queryId = INT64CONST(0);
+		/* See pgsm_ProcessUtility for more details regarding this skip logic */
+		if (!pgsm_track_utility ||
+			IsA(query->utilityStmt, PrepareStmt) ||
+			IsA(query->utilityStmt, DeallocateStmt))
+			return;
 
-		return;
+		if (IsA(query->utilityStmt, ExecuteStmt))
+		{
+			query->queryId = INT64CONST(0);
+			return;
+		}
 	}
 
 	/*
@@ -504,6 +511,16 @@ pgsm_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 											   location,	/* query location */
 											   &norm_query_len);
 	}
+
+	/*
+	 * This is optimization. The only thing that we need here for utility
+	 * statements is normalized query. If we don't have it, we can skip entry
+	 * creation. It will be created in pgsm_ProcessUtility() if needed.
+	 */
+	if (query->utilityStmt && norm_query == NULL)
+		return;
+
+	Assert(query->queryId != INT64CONST(0));
 
 	/*
 	 * pgsm_query_id always groups by the normalized form when we have one.
@@ -1101,8 +1118,6 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
 		!IsA(parsetree, PrepareStmt) &&
 		!IsA(parsetree, DeallocateStmt))
 	{
-		const char *query_text;
-		char	   *store_text;
 		int			location = pstmt->stmt_location;
 		int			query_len = pstmt->stmt_len;
 		int			cmd_type = pstmt->commandType;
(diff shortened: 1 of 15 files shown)


src/pg_stat_monitor.c
```diff
@@ -478,17 +478,23 @@ pgsm_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 	 */
 	if (query->utilityStmt)
 	{
-		if (pgsm_track_utility && IsA(query->utilityStmt, ExecuteStmt))
-			query->queryId = INT64CONST(0);
+		if (!pgsm_track_utility ||
+			IsA(query->utilityStmt, PrepareStmt) ||
+			IsA(query->utilityStmt, DeallocateStmt))
+			return;
```
jeltz, line 472: Why do we need this? This is not there in PGSS.

src/pg_stat_monitor.c
```diff
@@ -478,17 +478,23 @@ pgsm_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
...
+		}
 	}
 
 	/*
 	 * If we are unlucky enough to get a hash of zero, use 1 instead, to
 	 * prevent confusion with the utility-statement case.
 	 */
-	if (query->queryId == INT64CONST(0))
+	if (query->queryId == INT64CONST(0) && !query->utilityStmt)
```
jeltz: This code looks like a leftpver from PG 13 support. But probably best fixed as a separate PR.

src/pg_stat_monitor.c
```diff
@@ -512,6 +518,14 @@ pgsm_post_parse_analyze(ParseState *pstate, Query *query, JumbleState *jstate)
 											   &norm_query_len);
 	}
 
+	/*
+	 * This is optimization. The only thing that we need here for utility
+	 * statements is normalized query. If we don't have it, we can skip entry
+	 * creation. It will be created in pgsm_ProcessUtility() if needed.
+	 */
+	if (query->utilityStmt && norm_query == NULL)
+		return;
```
jeltz, line 508: Add an assert the queryid cannot be zero after this point? Or am I overly paranoid?

src/pg_stat_monitor.c
jeltz: Also can we even reach this is if has been assigned to zero? So i am unclear if this change would have been correct in PG 13.

src/pg_stat_monitor.c
```diff
@@ -1123,16 +1135,41 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+			 * we need to know this data at the statement execution start time.
+			 */
+			pgsm_fill_query_exec_info(&info);
+			query_text = CleanQuerytext(queryString, &location, &query_len);
+
+			pgsm_fill_query_stats(&stats, &info, queryId, 0,
+								  get_pgsm_query_id_hash(query_text, query_len),
+								  pnstrdup(query_text, query_len),	/* null terminated */
+								  cmd_type);
+		}
```
jeltz, line 1145: Is there a reason why you moved this before instead oif after? that slightly chsanges perforamce characteristics. Probably fine fir utility though, hmm.

src/pg_stat_monitor.c
```diff
@@ -1123,16 +1135,41 @@ pgsm_ProcessUtility(PlannedStmt *pstmt, const char *queryString,
...
+		else
+		{
+			pgsmQueryExecInfo info;
+			const char *query_text;
+
+			/*
+			 * Create the entry before utility statement execution because the
+			 * statement may change the execution info itself and for statistics
+			 * we need to know this data at the statement execution start time.
+			 */
```
jeltz: Is the last part actually true?

jeltz (changes requested):

src/pg_stat_monitor.c
artemgavrilov: For the scope of this PR: because we ignore these types of statements in `pgsm_ProcessUtility`, so nothing will consume this entries. But in general there is some specific rules how PGSM tracks these statements, see comments in the link below. Probably we should review it later.

https://github.com/percona/pg_stat_monitor/blob/01477543af6154213d835745ed45e56e0cdfb835/src/pg_stat_monitor.c#L1081-L1084

src/pg_stat_monitor.c
artemgavrilov: Added, yes there should not be 0 here

src/pg_stat_monitor.c
artemgavrilov: Yes, statement itself may change data that we collect (application name, user) and we need to collect it before execution. Before this PR we collected it in `pgsm_fill_query_exec_info` but I merged this func into `pgsm_fill_query_stats` as there no more reasons to have them separate.

src/pg_stat_monitor.c
artemgavrilov: Yes, when we execute `SET application_name = abc` we want to log app that executed this query, not `abc`

src/pg_stat_monitor.c
artemgavrilov: But I moved `getrusage/bufusage_start/walusage_start` below

src/pg_stat_monitor.c
artemgavrilov: Mmm, yes, it's leftover.

src/pg_stat_monitor.c
jeltz: But isnt that "execution info" and not "statistics"?

src/pg_stat_monitor.c
jeltz: Created a PR for this.

src/pg_stat_monitor.c
jeltz: Maybe add a comment here.

src/pg_stat_monitor.c
artemgavrilov: This comment, as well as separation to exec info and statistic has gone in the following commit.

src/pg_stat_monitor.c
artemgavrilov: Added

jeltz (approved):

PR #834, opened 2026-08-25
PG-2624 Hide sensitive data for users with no priveleges

artemgavrilov:
pg_stat_monitor exposes some data that can be considered as sensitive, so apply the same priveleges mechanism as for query text.

PG-2624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,3 +77,4 @@
 - Fix performance issue in comment extraction in large queries ([PG-1674](https://perconadev.atlassian.net/browse/PG-1674))
 - Fix error levels in PostreSQL 17 ([PG-1313](https://perconadev.atlassian.net/browse/PG-1313))
 - Fix bug where `cmd_type` often was 0 ([PG-1621](https://perconadev.atlassian.net/browse/PG-1621))
+- Hide sensitive date for users who has no priveleges ([PG-2624](https://perconadev.atlassian.net/browse/PG-2624)).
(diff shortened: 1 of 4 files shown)


src/pg_stat_monitor.c
```diff
@@ -2226,62 +2227,69 @@ pg_stat_monitor_internal(FunctionCallInfo fcinfo,
...
-		if (planid)
+		if (visible && planid)
 			values[i++] = Int64GetDatum(planid);
 		else
 			nulls[i++] = true;
 
-		if (may_read_all_stats || userid == GetUserId())
+		if (!showtext)
 		{
-			if (showtext)
```
artemgavrilov, line 2249: We have this `showtext` in the code, but don't expose it as parameter in API. I kept logic, but we have to discuss if we need it.

jeltz: well, it is kinda exposed in the API so changing this may count as an API breakage.

CHANGELOG.md
```diff
@@ -77,3 +77,4 @@
 - Fix performance issue in comment extraction in large queries ([PG-1674](https://perconadev.atlassian.net/browse/PG-1674))
 - Fix error levels in PostreSQL 17 ([PG-1313](https://perconadev.atlassian.net/browse/PG-1313))
 - Fix bug where `cmd_type` often was 0 ([PG-1621](https://perconadev.atlassian.net/browse/PG-1621))
+- Hide sensitive date for users who has no priveleges ([PG-2624](https://perconadev.atlassian.net/browse/PG-2624)).
```
jeltz, line 80: date -> data

jeltz (changes requested):

jeltz (approved):

PR #840, opened 2026-09-02
Ignore statements that have no query id

artemgavrilov:
Match pg_stat_statements behavior and ignore statements without query id. Otherwise all such statements stats will be merged into single entry as query id part of the hash key.

PG-0

### DIFF
--- a/regression/expected/level_tracking_4.out
+++ b/regression/expected/level_tracking_4.out
@@ -100,12 +100,11 @@ SELECT pg_stat_monitor_reset();
 CALL proc_with_utility_stmt();
 SELECT toplevel, calls, query FROM pg_stat_monitor
   ORDER BY query COLLATE "C", toplevel;
- toplevel | calls |              query              
-----------+-------+---------------------------------
+ toplevel | calls |             query              
+----------+-------+--------------------------------
  t        |     1 | CALL proc_with_utility_stmt()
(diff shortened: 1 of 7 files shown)


regression/expected/level_tracking_4.out
```diff
@@ -100,12 +100,11 @@ SELECT pg_stat_monitor_reset();
 CALL proc_with_utility_stmt();
 SELECT toplevel, calls, query FROM pg_stat_monitor
   ORDER BY query COLLATE "C", toplevel;
- toplevel | calls |              query              
-----------+-------+---------------------------------
+ toplevel | calls |             query              
+----------+-------+--------------------------------
  t        |     1 | CALL proc_with_utility_stmt()
  t        |     1 | SELECT pg_stat_monitor_reset()
- f        |     3 | SHOW pg_stat_monitor.pgsm_track
```
artemgavrilov, line 107: This was bogus stat. In fact it's 2+1 different statements, but because their query id was 0 they landed in one item. Now me match PGSS output. In PG16 query id for utility statements was fixed, so other output versions are fine.

jeltz (approved):
Do you want this in 2.4.0?

artemgavrilov:
> Do you want this in 2.4.0?

I think we can keep it for the next release. It's not good, but not critical.

PR #843, opened 2026-09-08
Merge back release 2.4.0

artemgavrilov:
PG-0


jeltz (approved):
Approved but not sure we actually need to create Prs for stuff like this.

artemgavrilov:
> Approved but not sure we actually need to create Prs for stuff like this.

Main branch is protected, we cannot push there (and I think it's good restriction)

PR #844, opened 2026-09-09
update packaging for PGSM-2.4.0

Manika-Percona:
PG-2746 https://perconadev.atlassian.net/browse/PG-2746


jeltz (changes requested):
The changes seem to affect more things than just 2.4.0 stuff so there should be an explanation but why these changes need to happen in the commit message and ideally it should be multiple commits,

